PR #1, opened 2013-02-01
Created fanClub project

dgreenhalgh:
Mostly placeholders and default values except where we've discussed.


bryansills:
Looks good.

PR #5, opened 2013-02-01
Refactored folders and started message structure

theJenix:
Messenger deserialization completed with unit test
added fanClub-test project and moved main app to sub folder called fanClub


dgreenhalgh:
Good idea, I like it

PR #9, opened 2013-02-06
4 spaces instead of tabs

reidbaker:
Sills didnt like mixing tabs and spaces this should fix that.


dgreenhalgh:
Probably a result of me writing this code on a fresh OS install on my desktop with Eclipse default values. Will fix on my end so it doesn't happen again.

PR #12, opened 2013-02-13
Monkey runner tests

reidbaker:


reidbaker:
I forgot to add to the read me will do before pulling this in

PR #16, opened 2013-02-18
Added bash script to rename APK file to last git commit hash

dgreenhalgh:


dgreenhalgh:
Waiting until we build to test this further

PR #17, opened 2013-02-18
Utilities folder and Song.java refactor

dgreenhalgh:
First two commits add a folder to hold scripts and other utilities, aptly named "utilities".

The third commit updates the script a bit for renameAPK.sh, but this is just a placeholder for now and won't be tested (read: finished) until we get the ability to test it on actual builds.
The third commit also (and more importantly) refactors song metadata out of Song.java and into SongMetadata.java. The distinction here is that SongMetadata contains everything that needs to be added to Playlist, and Song contains a SongMetadata object along with other fields.

The fourth commit adds the username field to SongMetadata, which I initially forgot.


utilities/renameAPK.sh
```diff
@@ -0,0 +1 @@
+mv ../signed-apks/fanClub.apk ../signed-apks/$(git log --pretty=format:'%h' -n 1).apk
```
reidbaker: Lets make this the same as test_results and have the folder be signed_apks. Also will you fix my gitignore mistakes so that everyone has the signed_apks folder

PR #26, opened 2013-02-27
Added playlist object

bryansills:
Added a class for the playlist. Also changed playlist in playlistfragment, etc. to list<songmetadata>

### DIFF
--- a/fanClub/src/com/lastcrusade/fanclub/model/Playlist.java
+++ b/fanClub/src/com/lastcrusade/fanclub/model/Playlist.java
@@ -1,0 +1,52 @@
+package com.lastcrusade.fanclub.model;
+
+import java.util.ArrayList;
+import java.util.List;
+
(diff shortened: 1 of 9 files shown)


fanClub/src/com/lastcrusade/fanclub/model/Playlist.java
```diff
@@ -0,0 +1,52 @@
...
+    }
+
+    public SongMetadata remove(int index) {
+        SongMetadata removeMeta = null;
+        if (index >= 0) {
+            removeMeta = list.remove(index);
+        }
+        return removeMeta;
+    }
+
```
dgreenhalgh, line 42: "Should add and remove to playlist return void here?" :horse:

reidbaker:
Can't merge this until project.properties is removed otherwise it looks good.

PR #31, opened 2013-02-27
Fixing the Titling Issues and Doing Some Refactoring

ejohnson44:
1. Got titling working in all cases except when you press back from the menu view
2. Renamed Titleable to ITitleable
3. Moved fragments into the components package
4. Cleaned up fragment code


fanClub/src/com/lastcrusade/fanclub/CoreActivity.java
```diff
@@ -43,80 +48,91 @@ public void onServiceDisconnected(ComponentName arg0) {
...
-        switchFragment(menuFragment, false);
-        
-        //setup the sliding bar
+        //add the initial content fragment and set the title on the action bar
+        switchActiveContent(getString(R.string.playlist));
+        setTitle(getString(R.string.playlist));
+
+       
+        // setup the sliding bar
         getSlidingMenu().setBehindOffset(60);
```
reidbaker: Magic number?

fanClub/src/com/lastcrusade/fanclub/CoreActivity.java
```diff
@@ -43,80 +48,91 @@ public void onServiceDisconnected(ComponentName arg0) {
...
-           setTitle(getString(R.string.app_name)); 
-           return true;
-   
-       }
-       
-       return false;
+        // home references the app icon
+        if (item.getItemId() == android.R.id.home) {
+            toggle(); // toggles the state of the sliding menu
+            if(getSlidingMenu().isMenuShowing() && menu.isAdded())
```
reidbaker: Could we get brackets around this if {}

fanClub/src/com/lastcrusade/fanclub/CoreActivity.java
```diff
@@ -43,80 +48,91 @@ public void onServiceDisconnected(ComponentName arg0) {
...
         }
+    }
+
+    /*
+     * Get the fragment associated with the name passed in.
+     * If later we decide we want to save the state of the fragments we can
+     * rework this to get saved states instead of creating a new fragment
+     * every time
+     */
+    private Fragment getFragment(String fragmentName) {
```
reidbaker, line 126: Why does this return null instead of being a void method?

fanClub/src/com/lastcrusade/fanclub/components/MusicLibraryFragment.java
```diff
@@ -10,12 +10,18 @@
...
+import com.lastcrusade.fanclub.R.string;
 import com.lastcrusade.fanclub.model.SongMetadata;
 import com.lastcrusade.fanclub.util.BluetoothUtils;
 import com.lastcrusade.fanclub.util.MusicListAdapter;
-import com.lastcrusade.fanclub.util.Titleable;
+import com.lastcrusade.fanclub.util.ITitleable;
 
-public class MusicLibraryFragment extends SherlockListFragment implements Titleable {
+public class MusicLibraryFragment extends SherlockListFragment implements ITitleable {
     private final String TAG = MusicLibraryFragment.class.getName();
```
reidbaker, line 25: Thanks for getting this

fanClub/src/com/lastcrusade/fanclub/components/PlaylistFragment.java
```diff
@@ -47,8 +49,9 @@ public View onCreateView(LayoutInflater inflater, ViewGroup container, Bundle sa
         return v;
     }
     
+    @Override
     public void onListItemClick(ListView lv, View v, int position, long id){
-        //code for expanding the particular song
+        //TODO: expanding the particular song
```
reidbaker, line 54: What this left on purpose?

fanClub/src/com/lastcrusade/fanclub/CoreActivity.java
ejohnson44: Good point - I'm not sure what exactly that will be yet - I don't like the size margin right now. I will comment about what it is and then play with it later

fanClub/src/com/lastcrusade/fanclub/CoreActivity.java
ejohnson44: Done.

fanClub/src/com/lastcrusade/fanclub/CoreActivity.java
ejohnson44, line 126: Because it returns the fragment - it only returns null if the fragment name does not match a fragment that we want to create

fanClub/src/com/lastcrusade/fanclub/components/PlaylistFragment.java
ejohnson44, line 54: Yes. Expanding the view is not crucial to anything we are currently working on or to getting the app actually functional, so I will get to this later

fanClub/src/com/lastcrusade/fanclub/CoreActivity.java
reidbaker: Lets call it `SIZE_MARGIN` for now. I think the proper way to play with those variables is to put them in a file called style.xml

fanClub/src/com/lastcrusade/fanclub/CoreActivity.java
reidbaker, line 126: Ahh I see. Could we set the fragment and have one return at the end? Also can you add brackets to these if's?

fanClub/src/com/lastcrusade/fanclub/util/BluetoothUtils.java
```diff
@@ -74,11 +75,20 @@ public static void enableDiscovery(Context context) {
...
+        if(mBluetoothAdapter!=null){
+            name = mBluetoothAdapter.getName();
+            if(name == null){
+                name = mBluetoothAdapter.getAddress();
+            }
+           
+        }
+        else{
+            Log.wtf(TAG, "No Bluetooth Radio. Ok on emulator");
+            name = "No Bluetooth";
```
reidbaker, line 88: We might want to put this in strings.xml for internationalization because we display this to the user. You dont have to but it should happen eventually.

PR #37, opened 2013-03-02
MusicLibraryFragment updating

ejohnson44:
Changed the distribution of work in updating the shown music library to reflect the state of the network


reidbaker:
:thumbsup: :moneybag: :monorail: :hatched_chick:

PR #41, opened 2013-03-08
Play things from queue

reidbaker:
This branch allows songs to be played from the playlist class and updates various ui components. Please checkout code and run on your device.

More importantly please check _small_ things like properly names variables, good comments for confusing code and formatting as well as _big_ things like architecture and the way information is passed.

### DIFF
--- a/fanClub/res/layout/content_frame.xml
+++ b/fanClub/res/layout/content_frame.xml
@@ -1,13 +1,15 @@
 <?xml version="1.0" encoding="utf-8"?>
-<FrameLayout xmlns:android="http://schemas.android.com/apk/res/android"
+<RelativeLayout xmlns:android="http://schemas.android.com/apk/res/android" >
+    
+    <FrameLayout
     android:id="@+id/content_frame"
     android:layout_width="match_parent"
-    android:layout_height="match_parent">
+    android:layout_height="match_parent"/>
     
     <FrameLayout
         android:id="@+id/playbar"
-        android:layout_gravity="bottom"
+        android:layout_alignParentBottom="true"
         android:layout_width="match_parent"
-        android:layout_height="wrap_content"
-    />
-</FrameLayout>
+        android:layout_height="wrap_content"/>
+
+</RelativeLayout>
--- a/fanClub/res/layout/content_frame.xml
+++ b/fanClub/res/layout/content_frame.xml
@@ -1,15 +1,17 @@
 <?xml version="1.0" encoding="utf-8"?>
 <RelativeLayout xmlns:android="http://schemas.android.com/apk/res/android" >
-    
+
     <FrameLayout
     android:id="@+id/content_frame"
     android:layout_width="match_parent"
     android:layout_height="match_parent"/>
-    
+
     <FrameLayout
-        android:id="@+id/playbar"
-        android:layout_alignParentBottom="true"
-        android:layout_width="match_parent"
-        android:layout_height="wrap_content"/>
+    android:id="@+id/playbar"
+    android:layout_alignParentBottom="true"
+    android:layout_width="match_parent"
+    android:layout_height="wrap_content"/>
 
 </RelativeLayout>
+
+
--- a/fanClub/res/layout/content_frame.xml
+++ b/fanClub/res/layout/content_frame.xml
@@ -2,16 +2,14 @@
 <RelativeLayout xmlns:android="http://schemas.android.com/apk/res/android" >
 
     <FrameLayout
-    android:id="@+id/content_frame"
-    android:layout_width="match_parent"
-    android:layout_height="match_parent"/>
+        android:id="@+id/content_frame"
+        android:layout_width="match_parent"
+        android:layout_height="match_parent" />
 
     <FrameLayout
-    android:id="@+id/playbar"
-    android:layout_alignParentBottom="true"
-    android:layout_width="match_parent"
-    android:layout_height="wrap_content"/>
+        android:id="@+id/playbar"
+        android:layout_width="match_parent"
+        android:layout_height="wrap_content"
+        android:layout_alignParentBottom="true" />
 
 </RelativeLayout>
-
-
--- a/fanClub/res/values/error_strings.xml
+++ b/fanClub/res/values/error_strings.xml
@@ -1,0 +1,5 @@
+<?xml version="1.0" encoding="utf-8"?>
+<resources>
+    <string name="playlist_empty">Playlist is empty</string>
+    
+</resources>
--- a/fanClub/src/com/lastcrusade/fanclub/CoreActivity.java
+++ b/fanClub/src/com/lastcrusade/fanclub/CoreActivity.java
@@ -2,8 +2,6 @@ package com.lastcrusade.fanclub;
 
 import android.os.Bundle;
 import android.support.v4.app.Fragment;
-import android.util.Log;
-import android.view.MotionEvent;
 
 import com.actionbarsherlock.view.MenuItem;
 import com.lastcrusade.fanclub.components.MenuFragment;
--- a/fanClub/src/com/lastcrusade/fanclub/CoreActivity.java
+++ b/fanClub/src/com/lastcrusade/fanclub/CoreActivity.java
@@ -2,6 +2,8 @@ package com.lastcrusade.fanclub;
 
 import android.os.Bundle;
 import android.support.v4.app.Fragment;
+import android.util.Log;
+import android.view.MotionEvent;
 
 import com.actionbarsherlock.view.MenuItem;
 import com.lastcrusade.fanclub.components.MenuFragment;
@@ -16,6 +18,7 @@ import com.slidingmenu.lib.app.SlidingFragmentActivity;
 public class CoreActivity extends SlidingFragmentActivity{
     private Fragment activeContent;
     private Fragment menu;
+    private final String TAG = CoreActivity.class.getName();
         
     public void onCreate(Bundle savedInstanceState){
         super.onCreate(savedInstanceState);
--- a/fanClub/src/com/lastcrusade/fanclub/CoreActivity.java
+++ b/fanClub/src/com/lastcrusade/fanclub/CoreActivity.java
@@ -55,7 +55,10 @@ public class CoreActivity extends SlidingFragmentActivity{
 
         //Add user to user list
         CustomApp curApp = ((CustomApp)getApplication());
-        curApp.getUserList().addUser(BluetoothUtils.getLocalBluetoothName());
+        
+        //TODO: Move this to something like connect activity or the connection fragment
+        curApp.getUserList().addUser(BluetoothUtils.getLocalBluetoothName(), BluetoothUtils.getLocalBluetoothMAC());
+        //Log.i("Core", " " + curApp.getUserList().getUserByMACAddress(BluetoothUtils.getLocalBluetoothMAC()));
     }
 
     public boolean onOptionsItemSelected(MenuItem item) {
--- a/fanClub/src/com/lastcrusade/fanclub/audio/SingleFileAudioPlayer.java
+++ b/fanClub/src/com/lastcrusade/fanclub/audio/SingleFileAudioPlayer.java
@@ -48,6 +48,7 @@ public class SingleFileAudioPlayer implements IPlayer {
 
     public void play() {
         try {
+            //This will fail and throw and Exception if the filepath is bad
             new File((new File(this.filePath).getParentFile().list())[0])
                     .exists();
             if (player.isPlaying()) {
--- a/fanClub/src/com/lastcrusade/fanclub/audio/SingleFileAudioPlayer.java
+++ b/fanClub/src/com/lastcrusade/fanclub/audio/SingleFileAudioPlayer.java
@@ -3,8 +3,10 @@ package com.lastcrusade.fanclub.audio;
 import java.io.File;
 
 import com.lastcrusade.fanclub.service.MessagingService;
+import com.lastcrusade.fanclub.service.PlaylistService;
 import com.lastcrusade.fanclub.util.BroadcastIntent;
 
+import android.content.Context;
 import android.media.MediaPlayer;
 import android.media.MediaPlayer.OnCompletionListener;
 import android.util.Log;
@@ -25,9 +27,14 @@ public class SingleFileAudioPlayer implements IPlayer {
     private String filePath;
     private MediaPlayer player;
 
-    public SingleFileAudioPlayer(OnCompletionListener onCompletion) {
+    public SingleFileAudioPlayer(final Context context) {
         this.player = new MediaPlayer();
-        player.setOnCompletionListener(onCompletion);
+        player.setOnCompletionListener(
+                new OnCompletionListener(){
+                    @Override public void onCompletion(MediaPlayer mp) {
+                        new BroadcastIntent(SingleFileAudioPlayer.ACTION_SONG_FINISHED).send(context);
+                    }
+        });
     }
 
     public void setSongByPath(String filePath) {
--- a/fanClub/src/com/lastcrusade/fanclub/audio/SingleFileAudioPlayer.java
+++ b/fanClub/src/com/lastcrusade/fanclub/audio/SingleFileAudioPlayer.java
@@ -2,28 +2,34 @@ package com.lastcrusade.fanclub.audio;
 
 import java.io.File;
 
+import com.lastcrusade.fanclub.service.MessagingService;
+import com.lastcrusade.fanclub.util.BroadcastIntent;
+
 import android.media.MediaPlayer;
+import android.media.MediaPlayer.OnCompletionListener;
 import android.util.Log;
 
-
 /**
- * A simple audio player that expects an audio file to be located in an accessible folder.
+ * A simple audio player that expects an audio file to be located in an
+ * accessible folder.
  * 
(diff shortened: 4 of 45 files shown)


fanClub/res/layout/content_frame.xml
```diff
@@ -1,17 +1,17 @@
...
-	    android:id="@+id/content_frame"
-	    android:layout_width="match_parent"
-	    android:layout_height="match_parent"/>   
-	
-	<FrameLayout
-        android:id="@+id/playbar"
-        android:layout_alignParentBottom="true"
-        android:layout_width="match_parent"
-        android:layout_height="wrap_content"/>
+    <FrameLayout
```
bryansills, line 4: You messed up the formatting on this.

fanClub/src/com/lastcrusade/fanclub/CoreActivity.java
```diff
@@ -2,6 +2,8 @@
 
 import android.os.Bundle;
 import android.support.v4.app.Fragment;
+import android.util.Log;
+import android.view.MotionEvent;
```
bryansills: looks like unnecessary imports

fanClub/src/com/lastcrusade/fanclub/util/MusicListAdapter.java
```diff
@@ -67,7 +67,7 @@ public View getView(int position, View convertView, ViewGroup parent) {
 
         addButton.setTag(position);
 
-        userColor.setBackgroundColor(Color.parseColor(users.get(metadataList.get(position).getUsername())));
+        userColor.setBackgroundColor(0xFFFFFFFF);
```
dgreenhalgh: Could this be not hardcoded?

fanClub/src/com/lastcrusade/fanclub/components/PlaybarFragment.java
```diff
@@ -34,23 +49,29 @@ public View onCreateView(LayoutInflater inflater, ViewGroup container,
...
+
+        playPause.setOnClickListener(new OnClickListener() {
+            @Override
+            public void onClick(View v) {
+                try {
+                    PlaylistService service = playlistServiceLocator.getService();
+                    if (service.isPlaying()) {
+                        service.pause();
+                        setPlayImage(playPause);
+                        Log.w(TAG, "pause called");
```
bryansills: should not be a warning

fanClub/src/com/lastcrusade/fanclub/service/PlaylistService.java
```diff
@@ -68,10 +64,19 @@ public PlaylistService getService() {
 
     private BroadcastRegistrar    registrar;
     private SingleFileAudioPlayer audioPlayer;
+    private Playlist queue;
```
dgreenhalgh: What's the reason for changing this variable name? Seems like @ejohnson44's evil doing

fanClub/src/com/lastcrusade/fanclub/components/PlaybarFragment.java
```diff
@@ -68,4 +89,30 @@ public void onClick(View v) {
 
         return view;
     }
+
+    private void setPlayImage(final ImageButton playPause) {
```
bryansills: why have this method if you just use the more verbose version when playing?

fanClub/src/com/lastcrusade/fanclub/service/PlaylistService.java
```diff
@@ -68,10 +64,19 @@ public PlaylistService getService() {
...
-        this.audioPlayer = new SingleFileAudioPlayer();
+        this.audioPlayer = new SingleFileAudioPlayer(
+                new OnCompletionListener(){
+                    @Override public void onCompletion(MediaPlayer mp) {
+                        new BroadcastIntent(SingleFileAudioPlayer.ACTION_SONG_FINISHED).send(PlaylistService.this);
+                    }
+        });
+
+        this.queue = new Playlist();
+        queue.add((new MediaStoreWrapper(this)).list().get(0));
```
bryansills: take out the test code.

fanClub/src/com/lastcrusade/fanclub/util/MusicListAdapter.java
bryansills: i think it was for testing. can we get back the original code?

fanClub/src/com/lastcrusade/fanclub/components/PlaybarFragment.java
```diff
@@ -34,23 +49,29 @@ public View onCreateView(LayoutInflater inflater, ViewGroup container,
...
-    
-                @Override
-                public void onClick(View v) {
-                    try {
-                        PlaylistService service = playlistServiceLocator.getService();
-                        if (service.isPlaying()) {
-                            service.pause();
-                        } else {
-                            service.play();
+        final ImageButton playPause = ((ImageButton) view.findViewById(R.id.btn_play_pause));
```
ejohnson44: why is the button final?

fanClub/src/com/lastcrusade/fanclub/components/PlaybarFragment.java
ejohnson44: The method name here doesn't make much sense to me. If you are setting the image, why does it take in the whole button? Why not just the image from the button or something like that? - actually, I see what is going on. But this leads to another question- why don't you just have this take in nothing and do the whole getView() and chaining inside here instead of outside?

fanClub/src/com/lastcrusade/fanclub/model/Playlist.java
```diff
@@ -47,4 +47,12 @@ public SongMetadata remove(int index) {
     public void setList(List<SongMetadata> list) {
         this.list = list;
     }
+    
+    public int size(){
+        return list.size();
+    }
+    
+    public SongMetadata getNext(){
+        return list.get(0);
```
ejohnson44: Doesn't this mean that we cannot loop through a playlist? I'm assuming that whenever a song is done playing it gets removed, but I thought we were just going to have a counter incremented or something so that we can get back to the previous songs if we have things repeating

fanClub/res/layout/content_frame.xml
reidbaker, line 4: fixed

fanClub/src/com/lastcrusade/fanclub/CoreActivity.java
reidbaker: fixed

fanClub/src/com/lastcrusade/fanclub/util/MusicListAdapter.java
reidbaker: fixed

fanClub/src/com/lastcrusade/fanclub/components/PlaybarFragment.java
reidbaker: fixed

fanClub/src/com/lastcrusade/fanclub/service/PlaylistService.java
reidbaker: It was my doing.  Do we not like name? If so what do you suggest?

fanClub/src/com/lastcrusade/fanclub/components/PlaybarFragment.java
reidbaker: fixed

fanClub/src/com/lastcrusade/fanclub/service/PlaylistService.java
reidbaker: this was the source of my null pointer exception code :( fixed

fanClub/src/com/lastcrusade/fanclub/components/PlaybarFragment.java
reidbaker: no longer final

fanClub/src/com/lastcrusade/fanclub/service/PlaylistService.java
dgreenhalgh: I just didn't know why it was changed. queue is likely more representative of the how it will be used, but playlist more closely aligns with the concept of a music playlist. 

Could be a design decision that gets repeated down the road. Thoughts, anyone?

reidbaker: Changed to mPlaylist for member playlist.

fanClub/src/com/lastcrusade/fanclub/audio/SingleFileAudioPlayer.java
```diff
@@ -32,13 +38,15 @@ public void setSongByPath(String filePath) {
...
+
     public void play() {
         try {
-            new File((new File(this.filePath).getParentFile().list())[0]).exists();
+            new File((new File(this.filePath).getParentFile().list())[0])
+                    .exists();
             if (player.isPlaying()) {
                 player.stop();
             }
+            player.reset();
```
ejohnson44, line 57: Why are we resetting the player every time?

reidbaker, line 57: Because we have to or better phrased I am under the impression we have to.
You have to reset the player or there is a weird bug where the player crashes if you change the state near the end of the file.

fanClub/src/com/lastcrusade/fanclub/components/PlaylistFragment.java
```diff
@@ -114,4 +118,32 @@ protected PlaylistService getPlaylistService() {
...
+    private class PlayListAdapter extends MusicListAdapter {
+        public PlayListAdapter(
+                Context mContext,
+                List<SongMetadata> metadataList,
+                UserList users
+                ) {
+            super(mContext, metadataList, users);
+        }
+
+        public void updateMusic(List<SongMetadata> unplayed, List<SongMetadata> played) {
```
ejohnson44, line 131: The more I think about this, I think it makes more sense for us to just keep track of the index of the song we are currently playing (probably in playlist service?), and then just have the view reference that position and compare its current position to that position. That way we are not creating a bunch of new lists every time we hit a new song. And then perhaps songs that are not fully transferred can have some kind of flag attached to them, or they can be in some sort of parallel list.

fanClub/src/com/lastcrusade/fanclub/model/Playlist.java
```diff
@@ -35,16 +43,56 @@ public SongMetadata remove(SongMetadata meta) {
...
+    public void setPlaylist(List<SongMetadata> unPlayed, List<SongMetadata> played){
+        this.unPlayed = unPlayed;
+        this.played = played;
+    }
+    
+    public int size(){
+        return unPlayed.size();
+    }
+    
+    public SongMetadata getHead(){
```
ejohnson44, line 75: Since you have unplayed below played, this is a little confusing - maybe change it to getNextUnplayed()?

fanClub/src/com/lastcrusade/fanclub/model/Playlist.java
```diff
@@ -35,16 +43,56 @@ public SongMetadata remove(SongMetadata meta) {
...
+     * @return new top of unplayed list
+     */
+    public SongMetadata moveNext(){
+        if(unPlayed.size() > 0){
+            //Add top of the unPlayed list to played list
+            played.add(unPlayed.get(0));
+            //Remove the top of unplayed list
+            unPlayed.remove(0);
+        }
+        if(unPlayed.size() > 0){
```
ejohnson44: This is the same as the condition above... is this correct?

Actually, I think this is why your last song would never play.

reidbaker: No the unplayed list shrinks at line 89. There is most likely a cleaner way to structure this but I couldnt think of it.

fanClub/src/com/lastcrusade/fanclub/model/Playlist.java
reidbaker, line 75: Yes this is left over from before there were 2 lists. I want to wait to change this until we decide on if we are keeping a playlist index. I think that might be best right now.

fanClub/src/com/lastcrusade/fanclub/components/PlaylistFragment.java
reidbaker, line 131: The more I think about this the more I think you are right. Unless there are any huge objections I say after we pull this in and I will do the playlist refactor as its own branch.

ejohnson44, line 131: Sounds good to me :] I'd like to get this functionality pulled in before we work on the refactor

fanClub/src/com/lastcrusade/fanclub/model/Playlist.java
ejohnson44: I see that it shrinks, but I still think this might be the cause of the problem since you are removing the song from unplayed before getting it again. I think that if you change it to this:

``` java
if(unPlayed.size()>0){
    SongMetadata song = unPlayed.get(0);
    played.add(song);
    unPlayed.remove(0);
    return song;
}
```

Then it will fix your problem

fanClub/src/com/lastcrusade/fanclub/audio/SingleFileAudioPlayer.java
```diff
@@ -32,13 +38,15 @@ public void setSongByPath(String filePath) {
     public boolean isPlaying() {
         return player.isPlaying();
     }
-    
+
     public void play() {
         try {
-            new File((new File(this.filePath).getParentFile().list())[0]).exists();
+            new File((new File(this.filePath).getParentFile().list())[0])
```
theJenix, line 52: Why do we have to do all this?  Isn't the filePath sufficient?

fanClub/src/com/lastcrusade/fanclub/service/PlaylistService.java
```diff
@@ -3,37 +3,38 @@
...
+import com.lastcrusade.fanclub.util.Toaster;
+import com.lastcrusade.fanclub.R;
 
 /**
- * This service is responsible for holding the play queue, and feeding songs to
- * the Audio player service.
- * 
- * @author Jesse Rosalia
- * 
+ * This service is responsible for holding the play queue
```
theJenix: Removed half of the javadoc comment, and it seems to still apply.

fanClub/src/com/lastcrusade/fanclub/service/PlaylistService.java
```diff
@@ -68,19 +66,26 @@ public PlaylistService getService() {
...
     private SingleFileAudioPlayer audioPlayer;
+    private Playlist mPlaylist;
 
     @Override
     public IBinder onBind(Intent intent) {
-        this.audioPlayer = new SingleFileAudioPlayer();
+        this.audioPlayer = new SingleFileAudioPlayer(
+                new OnCompletionListener(){
+                    @Override public void onCompletion(MediaPlayer mp) {
+                        new BroadcastIntent(SingleFileAudioPlayer.ACTION_SONG_FINISHED).send(PlaylistService.this);
```
theJenix: It looks like we send this broadcast action, receive it several lines down, and the broadcast a different action.  Can we collapse that process and just broadcast the different action from here?

fanClub/res/values/error_strings.xml
```diff
@@ -0,0 +1,5 @@
```
bryansills, line 0: maybe rename this file errors.xml? it goes more with the naming conventions of the other xml files.

fanClub/src/com/lastcrusade/fanclub/service/PlaylistService.java
reidbaker: Audio player service no longer exists ...

fanClub/src/com/lastcrusade/fanclub/service/PlaylistService.java
reidbaker: I don't think we should. What is happening here is we are passing an onCompleation listener to single file audio player so that we can be informed when the song is over. When the song is over the playlist service moves to the next song and broadcasts that the playlist is updated. The playlist fragment listens to playlist updated. 

In addition to playlist service the playbar fragment listens to song finished and changes the pause play button.

Basically I think it makes sense for single file audio player to say the song is finished and the playlist service to say the playlist has changed.

fanClub/res/values/error_strings.xml
reidbaker, line 0: I thought about that but I wanted to say what it was expected to hold. Anyone else want to chime in?

fanClub/src/com/lastcrusade/fanclub/service/PlaylistService.java
theJenix: Ahh good point.  But it still feeds songs to the SIngleFileAudioPlayer, so we can update the comment to reflect that.

fanClub/src/com/lastcrusade/fanclub/model/Playlist.java
reidbaker: The problem with not being able to play the last song was because I had `> 1` in some if not `> 0` or  `>=1`.  The problem with your code is I was intending to return the _new_ head of unPlayed not the old head. You shouldn't do that with out checking the _new_ length. 
I have looked at how the code is used however and I never take advantage of the returned value so I think I will change the return type to void.

PR #43, opened 2013-03-12
Network fragment

ejohnson44:
Created the network fragments - started to pull the network code into it, but realized that Jesse is the best person for this.

Also added some small feedback to adding songs to the queue. More feedback with be added later.


fanClub/res/layout/fragment_network.xml
```diff
@@ -0,0 +1,33 @@
+<?xml version="1.0" encoding="utf-8"?>
+<RelativeLayout xmlns:android="http://schemas.android.com/apk/res/android"
+    android:layout_width="match_parent"
+    android:layout_height="wrap_content" >
+
+    <Button 
+        android:id="@+id/add_members"
+        android:text="@string/add_members"
+        android:layout_width="match_parent"
+        android:layout_height="wrap_content"
```
reidbaker: unimportant nitpick: Should we close on the same line or the line after?

reidbaker:
I give it a :thumbsup: to pull in. Next person that agrees please pull it in.

fanClub/res/layout/fragment_network.xml
ejohnson44: Same line is probably best. I'll go in and fix it

fanClub/res/layout/user_item.xml
```diff
@@ -6,19 +6,20 @@
...
-        android:layout_width="12sp"
-        android:layout_height="12sp"
+        android:layout_width="20sp"
+        android:layout_height="20sp"
         android:layout_margin="4sp">
     </View>
 
     <TextView
-        android:id="@+id/username"
+        android:id="@+id/bluetoothID"
```
theJenix, line 15: Very small nit...we use snake case in all other ids.  I don't think this will prevent this from being pulled in, but it's something we should fix in a future commit.

theJenix:
This is good to go :+1:

PR #44, opened 2013-03-13
refactor and looping

reidbaker:
UI might need some polish from @ejohnson44 but it works

### DIFF
--- a/fanClub/src/com/lastcrusade/fanclub/model/Playlist.java
+++ b/fanClub/src/com/lastcrusade/fanclub/model/Playlist.java
@@ -7,92 +7,52 @@ import android.util.Log;
 
 public class Playlist {
     private final static String TAG = Playlist.class.getName();
-    private List<SongMetadata> unPlayed;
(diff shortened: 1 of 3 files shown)


dgreenhalgh:
I like it. Who wrote this code before it was refactored? It may be useful to have that person take a look at it.

reidbaker:
Me :)

dgreenhalgh:
Hahaha

fanClub/src/com/lastcrusade/fanclub/model/Playlist.java
```diff
@@ -7,92 +7,52 @@
...
         return removeMeta;
     }
 
-    public List<SongMetadata> getUnPlayedList() {
-        return unPlayed;
-    }
-    
-    public List<SongMetadata> getPlayedList(){
-        return played;
+    public List<SongMetadata> getSongsToPlay(){
```
bryansills, line 35: is this the best name for this method? i feel like with this name we should only get the unplayed songs.

PR #46, opened 2013-03-15
Newrelic

reidbaker:
Instructions and files needed to include new relics analytic's WOOHOO


fanClub/assets/newrelic.properties
```diff
@@ -0,0 +1,6 @@
+#
+#Thu Mar 14 20:43:23 EDT 2013
+new_relic_enable_ssl=true
+new_relic_enable_location=false
+new_relic_application_token=AAc68955bc22bc6ef5f6131e92e2552ef915f629a8
```
dgreenhalgh, line 5: :-1:

PR #49, opened 2013-03-15
Fix er up

theJenix:


fanClub/src/com/lastcrusade/fanclub/components/MusicLibraryFragment.java
```diff
@@ -38,7 +38,7 @@
     private boolean boundToService = false; //Since you cannot instantly bind, set a boolean
                                     // after its safe to call methods
     
-    private MusicListAdapter musicListAdapter;
+    private MusicAdapter musicAdapter;
```
reidbaker: how about mMusicAdapter

fanClub/src/com/lastcrusade/fanclub/service/PlaylistService.java
```diff
@@ -64,12 +66,16 @@ public PlaylistService getService() {
     }
 
     private BroadcastRegistrar    registrar;
+    private IPlayer               thePlayer;
```
dgreenhalgh, line 69: What kind of falcon is this?

PR #52, opened 2013-03-15
Add Song to Playlist Test

bryansills:
blah


ejohnson44:
Looks good!

addsongtoplaylisttest.py
```diff
@@ -0,0 +1,90 @@
+#! /usr/bin/env monkeyrunner
+'''
+Copyright (C) 2012  Diego Torres Milano
```
reidbaker, line 3: We cannot pull this code in unless we get the authors permission or find his licence. This is a blocking issue.

addsongtoplaylisttest.py
```diff
@@ -0,0 +1,90 @@
+#! /usr/bin/env monkeyrunner
```
reidbaker, line 1: This looks like a lot of code we dont need. After we figure out the copyright issue if we can use it lets pair it down to just what we are using.

dgreenhalgh:
Yeah, Sills, this is all around not okay.

I told Chip that we had these tests in place based on trust that these would be done. Please let me know when you can have this done, which needs to be ASAP because we're already past that deadline.

PR #54, opened 2013-03-15
Fanclub refactor two

reidbaker:


SoundStream/res/values/strings.xml
```diff
@@ -2,10 +2,10 @@
 <resources>
 
     <string name="app_uuid">e271f460-d1ad-4ea8-b555-5e59b19c71ff</string>
-    <string name="app_name">fanClub</string>
+    <string name="app_name">SoundStream</string>
     
     <!-- Landing Page Things -->
```
ejohnson44: You should be able to delete all these without causing any problems since you deleted landing activity. Other than that this all looks good :]

PR #55, opened 2013-03-16
Fixed bug where the host add members was not reenabled after no fans found

theJenix:


SoundStream/src/com/lastcrusade/soundstream/components/NetworkFragment.java
```diff
@@ -98,27 +98,37 @@ private void unregisterReceivers() {
...
+                            new IOnDialogItemClickListener<BluetoothDevice>() {
+    
+                                @Override
+                                public void onItemClick(
+                                        BluetoothDevice device) {
+                                    getConnectionService().connectToFan(device);
+                                }
+                            })
+                    .setFormatter(new BluetoothDeviceDialogFormatter())
+                    .show();
```
dgreenhalgh, line 130: Because now a competitor can see our code, we should probably one-line this to confuse them.

dgreenhalgh:
:+1:

reidbaker:
:ship:

PR #56, opened 2013-03-16
Menu Connected Users

ejohnson44:
the connected users list in menu now updates correctly


SoundStream/src/com/lastcrusade/soundstream/components/MenuFragment.java
```diff
@@ -54,15 +67,43 @@ public void onClick(View v) {
...
+     *
+     */
+    private void registerReceivers() {
+        this.registrar = new BroadcastRegistrar();
+        this.registrar.addAction(UserList.ACTION_USER_LIST_UPDATE, new IBroadcastActionHandler() {
+
+            @Override
+            public void onReceiveAction(Context context, Intent intent) {
+                //Update library shown when the library service gets an update
+                Log.i("MenuFragment", "User list update intent caught");
```
reidbaker: This should be a variable TAG. TAG should be `private static final String TAG = MenuFragment.class.getName()`

PR #62, opened 2013-03-17
Fan to host library

theJenix:
small refactor to move "on connection" code for fan into custom app, to be consistent with on connection code for host.


reidbaker:
:thumbsup: Next person that likes this pull it in.

ejohnson44:
:+1:

PR #64, opened 2013-03-17
Continuous play / title in playbar fragment

theJenix:
This was foundational, before I could tackle transfer song.  I also fixed the pause/resume functionality so it doesnt restart the song, and added a broadcast intent for the current song, and placed the title in the playbar fragment.  We can further style it however we like.

### DIFF
--- a/SoundStream/src/com/lastcrusade/soundstream/audio/RemoteAudioPlayer.java
+++ b/SoundStream/src/com/lastcrusade/soundstream/audio/RemoteAudioPlayer.java
@@ -13,6 +13,7 @@ import com.lastcrusade.soundstream.util.IBroadcastActionHandler;
 public class RemoteAudioPlayer implements IPlayer {
 
     private CustomApp application;
+    private boolean paused;
     private boolean playing;
     
     BroadcastRegistrar registrar;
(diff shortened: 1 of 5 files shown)


SoundStream/src/com/lastcrusade/soundstream/audio/RemoteAudioPlayer.java
```diff
@@ -57,9 +73,11 @@ private void registerReceivers() {
 			public void onReceiveAction(Context context, Intent intent) {
 				playing = intent.getBooleanExtra(MessagingService.EXTRA_IS_PLAYING, false);
 				if(playing) {
+				    paused = false;
```
reidbaker, line 76: You can fix later but there is weird spacing here

reidbaker:
:thumbsup: Next person that likes it pull this in.

SoundStream/src/com/lastcrusade/soundstream/audio/SingleFileAudioPlayer.java
```diff
@@ -43,7 +45,7 @@ public void setSongByPath(String filePath) {
 
     @Override
     public boolean isPlaying() {
-        return player.isPlaying();
+        return player.isPlaying() && !paused;
```
ejohnson44, line 48: Is there a reason that we are looking at player's playing, but our own pause? Later it looks like player has a paused state as well, so should we be using both of those here?

dgreenhalgh:
:shipit:

SoundStream/src/com/lastcrusade/soundstream/audio/SingleFileAudioPlayer.java
reidbaker, line 48: Was this addressed ?

PR #65, opened 2013-03-17
GPL copyright statements

dgreenhalgh:
in order to be protected under the GPL, we need to add a copyright statement to the top of every source file in our project. 
I've added three files to docs. 
copyright.txt is our general copyright statement, 
copyrightJava.txt is a version of the copyright statement formatted into a Java-style comment,
and copyrightXML.txt is a version of the copyright statement formatted into an XML-style comment.

I plan on adding these to the top of every existing file using sed, but it's up to you guys to add a copyright statement to the top of all files created after these are put into place.

Let me know if there are any changes that you'd like me to make before they're added to the top of every file because it could be a pain to update them.

### DIFF
--- a/SoundStream/src/com/lastcrusade/soundstream/components/ConnectFragment.java
+++ b/SoundStream/src/com/lastcrusade/soundstream/components/ConnectFragment.java
@@ -39,10 +39,14 @@ public class ConnectFragment extends SherlockFragment implements ITitleable{
     private BroadcastRegistrar broadcastRegistrar;
     private Button connectButton;
     
+    private ServiceLocator<MusicLibraryService> musicLibraryLocator;
+    
     @Override
     public void onCreate(Bundle savedInstanceState) {
         super.onCreate(savedInstanceState);
         registerReceivers();
+        
+        musicLibraryLocator = new ServiceLocator<MusicLibraryService>(this.getActivity(), MusicLibraryService.class, MusicLibraryServiceBinder.class);
 
     }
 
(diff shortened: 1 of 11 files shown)


docs/copyrightJava.txt
```diff
@@ -0,0 +1,18 @@
+/**
```
ejohnson44, line 1: Do we want this as a javadoc? Or just as a regular comment?

dgreenhalgh, line 1: Good call

dgreenhalgh:
Motherfucker. Alright so I only meant to add a change to that one line and I accidentally pushed a bunch of commits intended to cleanup (delete) our unused import statements.

So now there's a bit more code to review.

SoundStream/src/com/lastcrusade/soundstream/components/ConnectFragment.java
```diff
@@ -39,10 +39,14 @@
     private BroadcastRegistrar broadcastRegistrar;
     private Button connectButton;
     
+    private ServiceLocator<MusicLibraryService> musicLibraryLocator;
```
ejohnson44, line 42: I don't think this stuff is supposed to be here anymore - i think that was what @theJenix removed earlier today

theJenix, line 42: @ejohnson44 is correct..this was moved to CustomApp.  See the the "fan to host library" pull request.

reidbaker:
Feature branches :wink:

dgreenhalgh:
I used one, just messed it up

reidbaker:
Sorry my bad then. :thumbsup:

dgreenhalgh:
See pull request #73

PR #66, opened 2013-03-17
Playbar ui bug fix

ejohnson44:
Fixed a weird layout thing with how long songs were being shown

### DIFF
--- a/SoundStream/src/com/lastcrusade/soundstream/components/ConnectFragment.java
+++ b/SoundStream/src/com/lastcrusade/soundstream/components/ConnectFragment.java
@@ -74,7 +74,6 @@ public class ConnectFragment extends SherlockFragment implements ITitleable{
     @Override
     public void onDestroy() {
         unregisterReceivers();
-        musicLibraryLocator.unbind();
         super.onDestroy();
     }
     
--- a/SoundStream/src/com/lastcrusade/soundstream/model/SongMetadata.java
+++ b/SoundStream/src/com/lastcrusade/soundstream/model/SongMetadata.java
@@ -1,11 +1,13 @@
 package com.lastcrusade.soundstream.model;
 
-import com.lastcrusade.soundstream.util.DefaultParcelableCreator;
+import java.util.Arrays;
 
 import android.os.Parcel;
 import android.os.Parcelable;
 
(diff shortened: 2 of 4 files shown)


SoundStream/src/com/lastcrusade/soundstream/components/ConnectFragment.java
```diff
@@ -74,7 +74,6 @@ public void onClick(View v) {
     @Override
     public void onDestroy() {
         unregisterReceivers();
-        musicLibraryLocator.unbind();
```
reidbaker, line 77: Why did we get rid of this?

SoundStream/src/com/lastcrusade/soundstream/model/SongMetadata.java
```diff
@@ -99,4 +101,27 @@ public void setMacAddress(String macAddress) {
...
     }
 
+
+    /*
+     * To aid in ordering music - compares music by
+     * artist, then album, then song title
+     */
+    @Override
+    public int compareTo(SongMetadata another) {
+        int artistComp = artist.compareTo(another.getArtist());
```
reidbaker, line 111: This is the normal string compare right?

reidbaker, line 111: Also this code looks like a duplicate of another branch you have with a pull request

SoundStream/src/com/lastcrusade/soundstream/components/ConnectFragment.java
ejohnson44, line 77: To unbreak things - it was left over from another commit that gave me compiler errors

SoundStream/src/com/lastcrusade/soundstream/model/SongMetadata.java
ejohnson44, line 111: Oops. I made the new branch off of this most likely, instead of off master.... Is there a way to fix that? And yes, it is the normal string compare

ejohnson44, line 111: The only file that I meant to change on this one was the xml. I'll just close this pull request and open another that is actually correct. Sorry about that guys :frowning:

PR #67, opened 2013-03-17
Order music library

ejohnson44:
Gives the music library a basic ordering.

This has NOT been tested with multiple phones - I am pushing it now because I won't have access to another phone until Wednesday, but if anyone wants to see if this works as it they are welcome to test it out. If not, I will just test on Wednesday or Thursday and update it then

### DIFF
--- a/SoundStream/src/com/lastcrusade/soundstream/model/SongMetadata.java
+++ b/SoundStream/src/com/lastcrusade/soundstream/model/SongMetadata.java
@@ -1,13 +1,11 @@
 package com.lastcrusade.soundstream.model;
 
-import java.util.Arrays;
-
 import android.os.Parcel;
 import android.os.Parcelable;
 
 import com.lastcrusade.soundstream.util.DefaultParcelableCreator;
 
-public class SongMetadata implements Parcelable, Comparable<SongMetadata> {
+public class SongMetadata implements Parcelable{
 
     //this is REQUIRED for Parcelable to work properly
     public static final Parcelable.Creator<SongMetadata> CREATOR = new DefaultParcelableCreator(SongMetadata.class);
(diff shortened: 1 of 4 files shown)


SoundStream/src/com/lastcrusade/soundstream/model/SongMetadata.java
```diff
@@ -99,4 +101,27 @@ public void setMacAddress(String macAddress) {
         this.macAddress = macAddress;
     }
 
+
+    /*
+     * To aid in ordering music - compares music by
+     * artist, then album, then song title
+     */
+    @Override
+    public int compareTo(SongMetadata another) {
```
theJenix: Generally, it's good practice to not implement Comparator in the actual model object, because it pushes business logic into the model objects.  It might be better to either create this as an anonymous inner class in MusicLibraryService, or create this as a separate class that's used in MusicLibraryService.

SoundStream/src/com/lastcrusade/soundstream/service/MusicLibraryService.java
```diff
@@ -114,8 +114,13 @@ private void unregisterReceivers() {
 
     public List<SongMetadata> getLibrary() {
         synchronized(metadataMutex) {
+            
+            //sorts the list before returning it - for now,
+            //order is simply alphabetical by artist, album, then song.
+            ArrayList<SongMetadata> music = new ArrayList<SongMetadata>(metadataList);
```
theJenix: For large libraries, sorting on each call to getLibrary may introduce a lot of latency.  At a minimum, I think we want to move the sort outside the synchronized block, but what might be better in the short and long term would be to sort in updateLibrary, because that is a relatively infrequent action.

SoundStream/src/com/lastcrusade/soundstream/model/SongMetadata.java
ejohnson44: What kind of separate class are you thinking of? I've never done much work with comparable things in a large project like this, so I really didn't know of any conventions or best practices or anything like that

SoundStream/src/com/lastcrusade/soundstream/service/MusicLibraryService.java
ejohnson44: Alright, once I work out the other problems I'll move it there

theJenix:
This is :money_with_wings: :ship: :horse:

PR #68, opened 2013-03-18
formatting and notifyDataSetChanged for userlist

reidbaker:
Updates that were not right after the merge.

### DIFF
--- a/SoundStream/src/com/lastcrusade/soundstream/components/MenuFragment.java
+++ b/SoundStream/src/com/lastcrusade/soundstream/components/MenuFragment.java
@@ -95,7 +95,9 @@ public class MenuFragment extends SherlockFragment implements ITitleable {
             @Override
             public void onReceiveAction(Context context, Intent intent) {
                 //Update library shown when the library service gets an update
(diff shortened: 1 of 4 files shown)


reidbaker:
Shame on someone for braking master.

ejohnson44:
All looks good to me. Next person who agrees can pull it in

SoundStream/src/com/lastcrusade/soundstream/components/MenuFragment.java
```diff
@@ -95,7 +95,9 @@ private void registerReceivers() {
             @Override
             public void onReceiveAction(Context context, Intent intent) {
                 //Update library shown when the library service gets an update
-                ((UserListAdapter)userView.getAdapter()).updateUsers( ((CustomApp)getActivity().getApplication()).getUserList() );
+                ((UserListAdapter)userView.getAdapter()).updateUsers(
```
theJenix, line 98: I think we should use a member variable to hold the adapter, rather than casting userView.getAdapter() to the appropriate adapter.  It makes for more deliberate code; if we change the adapter, we know we have to change the member variable, and then anything that doesnt work with the new adapter class.  I don't think this will hold up this commit, but something to consider for future work on this class.

PR #70, opened 2013-03-18
Music view ui fix

ejohnson44:
The song view now fades nicely when there is more to the title or artist or album than can be viewed


theJenix:
This looks good to me.  Next person who likes it can :ship: it

PR #71, opened 2013-03-21
added menu icons

ejohnson44:
I might end up changing around the icons later, but these will work fine for our initial launch


dgreenhalgh:
These are tough to see in GitHub, but I trust they look good in-app. One more :+1: and we're good to go

theJenix:
Looks good to me.  Same comment as David, but I like what I see on Github so far.  This :ship: has come in.

PR #74, opened 2013-03-23
Bug fixes

theJenix:
Two network bug fixes:

There is a bug when reconnecting a guest after a host has gone down and come back up...this lets the guest reconnect w/o reopening the app.  The fix is to cancel the server socket after a successful connection.

There is a bug when receiving a bunch of messages that was introduced when I added the blocking read call to Messenger.  Messages were getting stuck in the input buffer, and not processed until the next byte was received...as a result, all messages were received late.  The fix was to remove the do...while loop, and make sure we run thru the message check/processing after each received byte.


dgreenhalgh:
:+1:

bryansills:
looks good to me.

PR #73, opened 2013-03-22
GPL copyright statements and import cleanup

dgreenhalgh:
Two features here: the addition of copyright statements to our documentation, and a comprehensive cleanup deleting the import statements we haven't been using.

PR #75, opened 2013-03-23
Menu disable

ejohnson44:
The menu is now disabled while in the connect fragment


SoundStream/src/com/lastcrusade/soundstream/components/ConnectFragment.java
```diff
@@ -13,16 +10,11 @@
...
 import com.lastcrusade.soundstream.CoreActivity;
 import com.lastcrusade.soundstream.CustomApp;
-import com.lastcrusade.soundstream.model.SongMetadata;
+import com.lastcrusade.soundstream.R;
 import com.lastcrusade.soundstream.service.ConnectionService;
 import com.lastcrusade.soundstream.service.IMessagingService;
-import com.lastcrusade.soundstream.service.MusicLibraryService;
-import com.lastcrusade.soundstream.service.ServiceLocator;
-import com.lastcrusade.soundstream.service.ServiceNotBoundException;
-import com.lastcrusade.soundstream.service.MusicLibraryService.MusicLibraryServiceBinder;
```
dgreenhalgh, line 25: Does the removal of these import statements conflict with #73 ?

ejohnson44, line 25: If it does I will just handle it when #73 gets pulled in

dgreenhalgh:
Looks like this can be auto-merged, even after #73 was pulled in. Merging in now...

PR #76, opened 2013-03-23
Connect dialog fix

theJenix:
Fixed multiple connect dialogs by changing Transitions to reuse fragment objects.  The bug was caused by multiple registrations; onCreate was called
every time we switched back to the NetworkFragment, but onDestroy was not being
called because we held the fragment on the back stack.

@ejohnson44 please look at this to make sure I didn't break anything fundamental with our fragments.


bryansills:
looks fine to me, but i'll let @ejohnson44 pull this one.

ejohnson44:
This all looks fine to me. I don't think it should mess with anything anywhere else.

PR #78, opened 2013-03-24
Added bash script to add GPL statement to the top of every source file

dgreenhalgh:
Added newlines to the end of copyrightJava.txt and copyrightXML.txt to make things pretty. 

Wrote a bash script to add our GPL statement to the top of all of our source files.

I've tested the script on a dummy checkout of our source and it works. I'm putting this pull request up now to make sure I haven't missed any of our source files. Actually running the script and pulling in the changes will be left for another pull request.

### DIFF
--- a/docs/copyrightJava.txt
+++ b/docs/copyrightJava.txt
@@ -16,3 +16,4 @@
  * You should have received a copy of the GNU General Public License
  * along with SoundStream.  If not, see <http://www.gnu.org/licenses/>.
  */
+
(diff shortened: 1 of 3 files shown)


dgreenhalgh:
The script looks for files containing the string "Copyright 2013 The Last Crusade ContactLastCrusade@gmail.com" in our source files. If the string doesn't exist in the file, the GPL statement is added to the top of that file. If that string is found in the file, the file is skipped. This is so you can run the script whenever you add a new file to the source.

If you write a new file, it's your responsibility to run this script before you make a pull request. Code not protected under our license should not be pulled into our source.

bryansills:
looks fine to me

PR #79, opened 2013-03-24
Preselect known guests

theJenix:
modified discovery process to keep track of previously paired guests, and mark them as "known".

modified dialog box to preselect the known guests


reidbaker:
This looks good the only thing I am not seeing is where we save the list of known people to a file or db or some kind of persistent storage.

theJenix:
Hmm...I thought I added it to the pull request comment.  I wanted to start with this for "Alpha" and look at introducing a DB for beta.  This should be good enough for a grand majority of our alpha testers, and I want to give the DB implementation a bit of time to make sure it doesnt cause any adverse affects.

reidbaker:
In that case :thumbsup: Next person that likes it merge it in.

PR #80, opened 2013-03-24
Big file fix

theJenix:
Fix big file loading by writing the file data out to a temporary file before sending the TRANSFER_DATA_MESSAGE broadcast intent, and sending the temp file path instead of the raw data.


reidbaker:
Looks good :thumbsup:

PR #81, opened 2013-03-24
Playlist clear on rotate fix

reidbaker:
This fixes the bug where you lose the playlist when you rotate the screen.
The things in CustomApp are my changes. The other things are import clean up and spacing cleanup.


theJenix:
:sailboat:

PR #88, opened 2013-03-25
Many music librarys

reidbaker:
This removes some code that kept the librarys from syncing properly.

PR #91, opened 2013-03-25
Discoverable discount

reidbaker:
Make guests discoverable for 30 seconds instead of for 5 min


theJenix:
As part of this fix, can we also reenable the connect button on the ConnectFragment after 30 seconds?  We can't get feedback when discoverability is disabled unless we poll the Bluetooth scan mode, which is probably more correct, but I think its fine to just reenable the button after 30 seconds.

reidbaker:
So I started working on the playbar fragment on the guests having the correct song. It was way bigger than I was thinking so I got it to a stable place and when this gets pulled in I will make that its own branch

dgreenhalgh:
Conflicts arose after #88 was pulled in

reidbaker:
This can now be pulled in

dgreenhalgh:
:snowflake: :snowflake: :snowflake: :snowflake: :snowflake: :snowflake: :snowflake: :snowflake: :snowflake: :snowflake: :snowflake: :snowflake: :snowflake: :snowflake:

PR #94, opened 2013-03-27
Refactor to allow us to run service tests.

theJenix:
.Removed use of CustomApp as a repo for services.  Instead, the services are bound in CustomApp to establish lifecycle, and each fragment or service will also bind the service within its lifecycle to use it.


SoundStream/src/com/lastcrusade/soundstream/audio/RemoteAudioPlayer.java
```diff
@@ -2,24 +2,32 @@
...
+import com.lastcrusade.soundstream.service.ServiceLocator;
+import com.lastcrusade.soundstream.service.ServiceNotBoundException;
 import com.lastcrusade.soundstream.util.BroadcastIntent;
 import com.lastcrusade.soundstream.util.BroadcastRegistrar;
 import com.lastcrusade.soundstream.util.IBroadcastActionHandler;
 
 public class RemoteAudioPlayer implements IPlayer {
 
-    private CustomApp application;
+    private static final String TAG = RemoteAudioPlayer.class.getSimpleName();
```
reidbaker, line 19: We have been using getName(). What is the difference and which should we use?

SoundStream/src/com/lastcrusade/soundstream/audio/RemoteAudioPlayer.java
```diff
@@ -55,13 +72,21 @@ public void resume() {
         //TODO: see above
         this.playing = true;
         this.paused  = false;
-        //TODO: this should probably send a resume message
-        this.application.getMessagingService().sendPlayMessage();
+        try {
+            //TODO: this should probably send a resume message
+            this.messagingService.getService().sendPlayMessage();
+        } catch (ServiceNotBoundException e) {
+            Log.wtf(TAG, e);
```
reidbaker, line 79: I wish there was a better option than just logging. Like trying again or throwing the exception for the caller to attempt to rebind and try again.

ejohnson44:
Looks good to me

SoundStream/src/com/lastcrusade/soundstream/audio/RemoteAudioPlayer.java
theJenix, line 79: This is probably something we need to evaluate system wide.  We log a lot in these cases, and when the app is finished that's not going to be very helpful.  Additionally, I'm considering making ServiceNotBoundException a RuntimException since, in all usage cases I can think of, it is not recoverable.

SoundStream/src/com/lastcrusade/soundstream/audio/RemoteAudioPlayer.java
theJenix, line 19: I thought we've used getSimpleName...getName includes the package, this is just the classname.  I'll change this to be in line with the rest of the system.

bryansills:
looks good to me. merging.

PR #95, opened 2013-03-27
Lock screen remote controls

theJenix:
Implemented the ICS RemoteControlClient, which gives us lock screen music on ICS and above.

### DIFF
--- a/SoundStream/src/com/lastcrusade/soundstream/CustomApp.java
+++ b/SoundStream/src/com/lastcrusade/soundstream/CustomApp.java
@@ -91,7 +91,16 @@ public class CustomApp extends Application {
 
             @Override
             public void onAudioFocusChange(int focusChange) {
-                //TODO: duck audio or pause when focus has changed.
+                //TODO: duck audio or pause in other cases where focus has changed.
+                switch (focusChange) {
+                //handle loss of focus, which includes when a phonecall is coming in
+                case AudioManager.AUDIOFOCUS_LOSS:
+                case AudioManager.AUDIOFOCUS_LOSS_TRANSIENT:
+//                case AudioManager.AUDIOFOCUS_LOSS_TRANSIENT_CAN_DUCK:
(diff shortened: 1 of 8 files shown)


theJenix:
Added the pause on phone call fix here, because I :truck:ed up.

SoundStream/src/com/lastcrusade/soundstream/CustomApp.java
```diff
@@ -58,11 +69,47 @@ public void onCreate() {
...
+            @Override
+            public void onAudioFocusChange(int focusChange) {
+                //TODO: duck audio or pause in other cases where focus has changed.
+                switch (focusChange) {
+                //handle loss of focus, which includes when a phonecall is coming in
+                case AudioManager.AUDIOFOCUS_LOSS:
+                case AudioManager.AUDIOFOCUS_LOSS_TRANSIENT:
+//                case AudioManager.AUDIOFOCUS_LOSS_TRANSIENT_CAN_DUCK:
+                    //NOTE: this calls the playlist service directly, because we want instant action.
+                    getPlaylistService().pause();
```
bryansills, line 101: does this run under the AUDIOFOCUS_LOSS_TRANSIENT case or the _CAN_DUCK case? I think this code can be cleaned up.

theJenix, line 101: The _CAN_DUCK case is commented out...but probably shouldnt pause the song...rather it should probably just duck the music.  There's another card in Trello to implement all of those basic audio player niceties, which will clean this code up in the process.

bryansills:
What's the link that you based this work on? I feel like I could get a better understanding of the code if I saw it outside of our code.

reidbaker:
We can pull this in because it has the call fix. But I really dont think any more features should be added till after we push to the play store. There are a ton of bugs and with the addition of this cic video I think our Friday deadline will slip.

PR #102, opened 2013-03-29
Messenger version

theJenix:
Messenger version is inserted into the stream, and pulled out in deserialize.  It currently doesn't do anything in deserialize.


SoundStream/src/com/lastcrusade/soundstream/net/message/Messenger.java
```diff
@@ -28,10 +28,14 @@
 
     private static final String TAG = Messenger.class.getName();
 
-    private static final int SIZE_LEN = 4;
+    //package protected, so they can be accessed from the
```
reidbaker: So they can be accessed from the ... what?

reidbaker:
Other than my comment this looks good to me. :rabbit:

PR #103, opened 2013-03-29
Proofread string grammar

dgreenhalgh:
Cleaned up string typos, replaced hardcoded strings, and subsequently added to our eToast code

### DIFF
--- a/SoundStream/src/com/lastcrusade/soundstream/util/Toaster.java
+++ b/SoundStream/src/com/lastcrusade/soundstream/util/Toaster.java
@@ -36,8 +36,7 @@ public class Toaster {
     }
     
     public static void eToast(Context context, int resId) {
-    	Log.e(Thread.currentThread().getStackTrace()[methodDepth].toString(), context.getString(resId));
-    	toastWorker(context, context.getString(resId));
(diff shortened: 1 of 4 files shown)


reidbaker:
Looks perfect! Next person pull this in :octocat:

SoundStream/src/com/lastcrusade/soundstream/util/Toaster.java
```diff
@@ -34,5 +34,14 @@ public static void eToast(Context context, String s) {
         Log.e(Thread.currentThread().getStackTrace()[methodDepth].toString(), s);
         toastWorker(context, s);
     }
+    
+    public static void eToast(Context context, int resId) {
+    	Log.e(Thread.currentThread().getStackTrace()[methodDepth].toString(), context.getString(resId));
```
theJenix: This should call the eToast method that takes a string, and pass in context.getString(resId) as that string.  See iToast.

theJenix:
One small change required in Toaster.java and then it's good to :goat:

dgreenhalgh:
Good catch, this should be good to go

theJenix:
:shoe: in

PR #106, opened 2013-04-01
Connect button did not know the state of discoverability 

reidbaker:
Connect button in connect fragment reflects the state of the world


SoundStream/src/com/lastcrusade/soundstream/components/ConnectFragment.java
```diff
@@ -127,6 +121,25 @@ public void onReceiveAction(Context context, Intent intent) {
                         ((CoreActivity)getActivity()).enableSlidingMenu();
                     }
                 })
+             .addAction(BluetoothAdapter.ACTION_SCAN_MODE_CHANGED, new IBroadcastActionHandler() {
```
theJenix, line 124: Love it!  My only request is to use a switch statement below, for readability...otherwise, this is a solid change.

reidbaker:
Good call. I changed it.

ejohnson44:
Looks Great :]

PR #107, opened 2013-04-01
Transfer state on guests incorrect

reidbaker:
fix the issue where the last song on every guests playlist was being displayed as transfering


bryansills:
looks fine to me

theJenix:
Does this only solve the problem for local files?

ejohnson44:
If I remember correctly, it was only a problem with local files because of the way they went through the data manager and when in that process things were updated

PR #109, opened 2013-04-01
implement pause on headphone dissconnect

reidbaker:
This still doesn't have audio ducking on notifications but it does pause when the headphone is disconnected.


ejohnson44:
Looks good to me :lock: :key:

PR #110, opened 2013-04-01
Clickable area

ejohnson44:
increased the size of the clickable areas on the buttons


reidbaker:
So It looks good to me and by good I mean I have no idea how to code review xml changes. You are a wizard

dgreenhalgh:
I'm in the same boat as Reid, but I like this

reidbaker:
http://www.youtube.com/watch?feature=player_detailpage&v=WX_OW_IuF5A#t=8s

reidbaker:
The image does not get bigger I am pulling this in.

PR #111, opened 2013-04-01
Title naming on back

ejohnson44:
fixed the title issue and Reid's compiler errors

### DIFF
--- a/SoundStream/src/com/lastcrusade/soundstream/components/MusicListFragment.java
+++ b/SoundStream/src/com/lastcrusade/soundstream/components/MusicListFragment.java
@@ -19,10 +19,11 @@ public abstract class MusicListFragment extends SherlockListFragment implements 
     }
 
     @Override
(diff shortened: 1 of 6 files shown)


reidbaker:
fantastic! :ship: it.

SoundStream/src/com/lastcrusade/soundstream/components/MusicListFragment.java
```diff
@@ -19,10 +19,11 @@ public View onCreateView(LayoutInflater inflater, ViewGroup container, Bundle sa
...
 
     @Override
-    public abstract int getTitle();
-    
-    @Override
     public void onResume(){
         super.onResume();
+        getActivity().setTitle(getTitle());
     }
+
```
reidbaker, line 26: remove this if it is not needed

reidbaker:
This should be pulled in next so that we fix my build issue.

PR #112, opened 2013-04-01
band-aid fix for the rotation issue

bryansills:
This is not the correct way to fix this issue, but it will get us through to the alpha.

The correct way to handle rotation is to save the necessary data in an overridden onSaveInstanceState method and pull them out of the bundle in onCreate. However, if we try to do it this way right now, we don't have a Bundle when onCreate is called after the rotation.  The Bundle is created correctly but onCreate is called twice for some reason with the Bundle  missing in the second calling of onCreate.


reidbaker:
This is for the entire application not just the playbar fragment. Let me see if I can find a better way.

ejohnson44:
See #113

PR #113, opened 2013-04-01
removed the playbar from the connection fragment

ejohnson44:


SoundStream/src/com/lastcrusade/soundstream/CoreActivity.java
```diff
@@ -58,11 +60,7 @@ public void onCreate(Bundle savedInstanceState){
```
reidbaker, line 60: A bunch of lines of whitespace

reidbaker:
Added trello card for adding isbound to the service locator.

reidbaker:
We can fix the whitespace later I am going to merge this in.

PR #117, opened 2013-04-02
make master compile, rename png

reidbaker:
I would like for master to no longer be broken


bryansills:
looks good to me. i'm gonna pull it in so that reid and i can continue working.

PR #118, opened 2013-04-02
clear should also set the entryId to zero

reidbaker:
This caused the last entry to appear to be transferring for a while


dgreenhalgh:
:+1:

ejohnson44:
Yay for one line pull requests :money_with_wings:

PR #121, opened 2013-04-02
Disconnect and Disband

bryansills:
Look over this carefully Reid and I just finished this and it is 4:20 in the morning.


SoundStream/src/com/lastcrusade/soundstream/CoreActivity.java
```diff
@@ -138,4 +136,6 @@ public void showPlaybar(){
         .replace(R.id.playbar, playbar)
         .commit();
     }
+
+    //TODO: LizzieJohn make a hidePlaybar method
```
theJenix, line 140: Is there a trello card for this?

SoundStream/src/com/lastcrusade/soundstream/components/NetworkFragment.java
```diff
@@ -90,9 +107,66 @@ public void onClick(View v) {
         
         this.adapter = new UserListAdapter(getActivity(), ((CustomApp)getActivity().getApplication()).getUserList(), false );
         users.setAdapter(this.adapter);
+
+        disconnect = (Button)v.findViewById(R.id.disconnect_btn);
+        disband = (Button)v.findViewById(R.id.disband_btn);
+
+        setDisconnectDisbandVisibility(disconnect, disband);
```
theJenix, line 114: This is also done in the ConnectionService onBind method...why do it in both places?

SoundStream/src/com/lastcrusade/soundstream/components/NetworkFragment.java
```diff
@@ -133,6 +207,25 @@ public void onReceiveAction(Context context, Intent intent) {
...
+                @Override
+                public void onReceiveAction(Context context, Intent intent) {
+                    setDisconnectDisbandVisibility(disconnect, disband);
+                }
+            })
+            .addAction(ConnectionService.ACTION_HOST_DISCONNECTED, new IBroadcastActionHandler() {
+                @Override
+                public void onReceiveAction(Context context, Intent intent) {
+                    //after the host has been disconnected, wipe everything and start fresh
+                    disconnect();
```
theJenix, line 226: I don't like this calling disconnect, as we're already disconnected.  Recommend separating that method into the method that actually disconnects (called from the dialog) and a method called cleanUpAfterDisconnect or something.

SoundStream/src/com/lastcrusade/soundstream/components/NetworkFragment.java
```diff
@@ -140,6 +233,49 @@ private void unregisterReceivers() {
         this.broadcastRegistrar.unregister();
     }
 
+    private void disconnect() {
+        getConnectionService().disconnectHost();
+
+        playlistServiceLocator = new ServiceLocator<PlaylistService>(
```
theJenix, line 239: Possible refactor (not for now, but for after alpha): introduce a "one time call" function in ServiceLocator that handles binding, calling the code, and then unbinding itself.  I like this pattern, but this could be confusing.

SoundStream/src/com/lastcrusade/soundstream/service/MusicLibraryService.java
```diff
@@ -332,4 +336,9 @@ private IMessagingService getMessagingService() {
         }
         return messagingService;
     }
+
+    public void clearExternalMusic() {
+        metadataList = getMyModifiableLibrary();
```
theJenix, line 341: This does not update the metadataMap, and is kind of confusing.  Instead of getMyModifiableLibrary, I would suggest calling getMyLibrary(), clearing the library, and then calling updateLibrary with the results from getMyLibrary.  I think it's clear what that's doing.

SoundStream/src/com/lastcrusade/soundstream/service/PlaylistService.java
```diff
@@ -340,14 +342,18 @@ public void onReceiveAction(Context context, Intent intent) {
                 boolean played    = intent.getBooleanExtra(MessagingService.EXTRA_PLAYED, false);
 
                 SongMetadata song = getMusicLibraryService().lookupSongByAddressAndId(macAddress, songId);
```
theJenix, line 344: I know there's trello card for this, but I recommend removing song entirely here.  We have the entryId, that should uniquely identify an entry on the playlist, and it removes a ton of complexity.

SoundStream/src/com/lastcrusade/soundstream/components/NetworkFragment.java
```diff
@@ -140,6 +233,49 @@ private void unregisterReceivers() {
...
+        });
+
+        ((CustomApp)getActivity().getApplication()).clearExternalUsers();
+
+        Transitions.transitionToConnect((CoreActivity) getActivity());
+    }
+
+    private void disband() {
+        getConnectionService().disconnectAllGuests();
+        disconnect();
```
theJenix, line 276: See comment above about disconnect...this is confusing here, and should call that cleanUpAfterDisconnect method instead.

theJenix:
This looks pretty good...just a couple changes needed for clarity.  :cake:

SoundStream/src/com/lastcrusade/soundstream/CoreActivity.java
reidbaker, line 140: There is now

SoundStream/src/com/lastcrusade/soundstream/components/NetworkFragment.java
reidbaker, line 114: Because we only bind in onCreate. If things have changed and onCreateView gets called then we should set the proper buttons.

SoundStream/src/com/lastcrusade/soundstream/components/NetworkFragment.java
reidbaker, line 226: I agree. We can break this up into 2 methods

SoundStream/src/com/lastcrusade/soundstream/components/NetworkFragment.java
reidbaker, line 239: Made a trello card

SoundStream/src/com/lastcrusade/soundstream/components/NetworkFragment.java
theJenix, line 114: If possible, I would suggest we react to those things changing.  On initial display of this view (or if we change the fact that the services are bound in onCreate), the services may not be bound at this point.  A comment and trello card is ok to move on here, but I think we want to clean this up soon.

SoundStream/src/com/lastcrusade/soundstream/service/MusicLibraryService.java
reidbaker, line 341: We did that first. You get a error because getMyLiibrary returns an unmodifiable list and updateLibrary then attempts to modify the objects.

theJenix:
Actually, sorry...an update to my previous comment.  Most things look good, the change to MusicLibraryService will break things because you're not updating the metadataMap.  That needs to be addressed.

SoundStream/res/values/strings.xml
```diff
@@ -50,10 +50,16 @@
 	<string name="network">Network</string>
 	<string name="music_library">Music Library</string>
 	<string name="playlist">Playlist</string>
+    <string name="disconnect">Disconnect</string>
```
ejohnson44, line 53: weird spacing here - not super important, but just thought I would mention it

SoundStream/res/layout/fragment_network.xml
```diff
@@ -46,6 +45,27 @@
         android:layout_alignLeft="@+id/connected_users_label">
     </ListView>
 
-
-
+    <Button
```
ejohnson44, line 48: since you set visibility later anways, I would go ahead and start this button off as invisible like we do with all the other buttons

SoundStream/src/com/lastcrusade/soundstream/components/NetworkFragment.java
reidbaker, line 114: Trello card made, Comment added, I branched off this code and hope to fix the issues shortly.

SoundStream/src/com/lastcrusade/soundstream/components/NetworkFragment.java
reidbaker, line 226: broken up now

SoundStream/src/com/lastcrusade/soundstream/service/PlaylistService.java
reidbaker, line 344: @theJenix  would you be willing to add this spot as a comment on that card.

SoundStream/res/values/strings.xml
reidbaker, line 53: Think it is lined up in real life. It is offset because of the +

reidbaker:
Closing this because of new pull request.

PR #122, opened 2013-04-02
Disconnect disband

reidbaker:
This is a continuation of #121 but with all the comments addressed


reidbaker:
Ok we have one issue and I need help tracking it down.
- Make a network with a guest and fan.
- On guest navigate to network fragment press disconnect
- Watch as you get a service not bound exception in onServiceBound
- Bang head against computer

SoundStream/src/com/lastcrusade/soundstream/service/MusicLibraryService.java
```diff
@@ -331,4 +335,12 @@ private IMessagingService getMessagingService() {
         }
         return messagingService;
     }
+
+    public void clearExternalMusic() {
+        synchronized (metadataMutex) {
+            metadataList = getMyModifiableLibrary();
+            metadataMap.clear();
```
theJenix: You can't just clear the map...you need to rebuild it.  Take a look at updateLibrary for an example.

reidbaker: Why not. It works

reidbaker:
Once @theJenix gives this the ok could the next person pull it in. 
When this is pulled in Ill push to the play store

ejohnson44:
Looks good to me. Once @theJenix  OK's it I can pull it in if he can't for whatever reason

SoundStream/src/com/lastcrusade/soundstream/service/MusicLibraryService.java
```diff
@@ -331,4 +335,11 @@ private IMessagingService getMessagingService() {
         }
         return messagingService;
     }
+
+    public void clearExternalMusic() {
+        metadataList = getMyModifiableLibrary();
+        metadataMap.clear();
+        orderAlphabetically();
```
theJenix, line 342: I'm still not 100% thrilled with this, but I won't hold up the merge because of it.  Can you please add a trello refactor card (or a comment to the card that talks about pulling out the list/map datastructure) to revisit this function and look for a more clear/explicit way of doing this?

theJenix:
:cake:

PR #123, opened 2013-04-03
Gitignore update for new relic

reidbaker:
new relic updated their sdk this will let us ignore the rest of the versions of the sdk


dgreenhalgh:
:+1:

PR #127, opened 2013-04-08
Song loading indicator

ejohnson44:


dgreenhalgh:
:shipit:

bryansills:
how does this tie into the rest of the app? does it update as the song gets loaded?

ejohnson44:
Every time the state changes, the view gets updated, so all of this just happens in the getView method for the list; for every song element, it checks the state of the song and colors/sets visibility accordingly

SoundStream/src/com/lastcrusade/soundstream/components/PlaylistFragment.java
```diff
@@ -176,11 +176,13 @@ public View getView(int position, View convertView, ViewGroup parent) {
             PlaylistEntry entry = super.getItem(position);
             if (!entry.isLoaded()) {
                 //TODO: style the unloaded elements here
-                element.setBackgroundColor(getResources().getColor(R.color.loading));
+                element.setBackgroundColor(getResources().getColor(R.color.abs__bright_foreground_disabled_holo_light));
+                element.findViewById(R.id.progress).setVisibility(View.VISIBLE);
```
bryansills, line 189: it is my impression that just because a song is in the playlist does not mean it is currently being loaded. i think this creates a binary distinction between loaded and not loaded when a song could also be not loaded and not ready to be loaded.

ejohnson44, line 189: But if it is not loaded it is still not able to be played until it is loaded. The indicator is more to say that this particular song cannot be played yet, but it is in the line of things that will be downloaded. It would be very misleading to users to have a song be in a normal looking state, and then go into a loading state, rather than just be loading for a long-ish period of time

reidbaker:
@bryansills have all your issues been addressed?

SoundStream/res/layout/song_item.xml
```diff
@@ -108,5 +108,15 @@
         android:visibility="invisible"
         android:contentDescription="@string/remove_from_playlist"
         android:background="@color/transparent"/>
-    
+     
+     <ProgressBar 
+         android:id="@+id/progress"
+         android:layout_width="wrap_content"
+         android:layout_height="wrap_content"
+         android:paddingRight="16sp"
```
reidbaker, line 114: This should be dp

ejohnson44, line 114: I'm going to do another pull request later that fixes them all at once

PR #129, opened 2013-04-10
Connect fragment makeover

ejohnson44:
vert
horiz


reidbaker:
I like it.
3 quick questions:
1. Are the images clickable?
2. Can we move the text to the center?
3. Do you think the create should indicate that they could add people later?

bryansills:
I like the design, but I don't think this shows the user that the two options are buttons that can be pressed. I think if we incorporate the text and image into a button, and add a little bit of padding, it will be more clear.

Also, didn't we want to change the 'Join' text to 'Ask to Join'?

reidbaker:
@bryansills I think if we change the title bar to "Choose one" or "Select" or something that could help

ejohnson44:
@reidbaker 
1. Yes :]
2. I'll see what it looks like
3. I think the join image works well enough for that

@bryansills 
Yes, the phrasing needs to be changed some, and I will play around with some kind of padding type thing to make it look more clickable

theJenix:
:ship:

PR #130, opened 2013-04-10
Layout refactor

ejohnson44:
Replaced all of the sp's with dp's except with text sizes


reidbaker:
Looks good next person pull it in :diamond_shape_with_a_dot_inside:

dgreenhalgh:
:shipit:

PR #131, opened 2013-04-10
Creation of userlist service

reidbaker:
This moves all our code to the new userlist service. 
A side effect was that I made the musiclibarary fragment use a service locator for the music library service.

There is also code to null check some buttons that were crashing on rotation.


SoundStream/src/com/lastcrusade/soundstream/components/MenuFragment.java
```diff
@@ -130,4 +147,25 @@ public void onReceiveAction(Context context, Intent intent) {
     private void unregisterReceivers() {
         this.registrar.unregister();
     }
+
+    private UserList getUserListFromService(){
+        UserList activeUsers = new UserList();
```
ejohnson44: It seems like we make this variable for no real reason. We could just return a new UserList in the bottom return

SoundStream/src/com/lastcrusade/soundstream/components/MusicLibraryFragment.java
```diff
@@ -44,81 +45,72 @@
...
-            boundToService = false;
-        }
-    };
-
-
     @Override
     public void onCreate(Bundle savedInstanceState){
         super.onCreate(savedInstanceState);
 
+        //make a new music list adapter and give it an empty list of songs and empty list of users to use until the service is connected
```
ejohnson44: split this onto two lines?

SoundStream/src/com/lastcrusade/soundstream/components/MusicLibraryFragment.java
```diff
@@ -170,6 +161,49 @@ protected PlaylistService getPlaylistService() {
...
+        if(musicLibraryService != null){
+            library = new ArrayList<SongMetadata>(musicLibraryService.getLibrary());
+        } else {
+            Log.i(TAG, "MusicLibarysService null, returning empty library");
+        }
+        return library;
+    }
+
+
+    private UserList getUserListFromService(){
```
ejohnson44, line 200: Same as before, though you could also change it to do what the music library version does

SoundStream/src/com/lastcrusade/soundstream/components/NetworkFragment.java
```diff
@@ -82,6 +86,18 @@ public void onServiceBound() {
             }
         });
 
+        NetworkFragment.this.adapter = new UserListAdapter(getActivity(), new UserList(), false );
```
ejohnson44: Is there a reason that you include NetworkFragment before this.adapter?

SoundStream/src/com/lastcrusade/soundstream/components/ConnectFragment.java
```diff
@@ -156,19 +156,21 @@ public void onReceiveAction(Context context, Intent intent) {
...
-                    case BluetoothAdapter.SCAN_MODE_CONNECTABLE:
-                        connectButton.setEnabled(true);
-                        break;
-                    case BluetoothAdapter.SCAN_MODE_CONNECTABLE_DISCOVERABLE:
-                        connectButton.setEnabled(false);
-                        break;
-                    default:
-                        Log.wtf(TAG, "Recieved scan mode changed with unknown mode");
-                        break;
+                    if(connectButton != null){
```
theJenix, line 159: Maybe not part of this pull request, but we should make sure we're not "paused"...that way, we don't respond to these actions when performed outside the app.

PR #135, opened 2013-04-12
Add mac and bluetooth name to UserListservice

reidbaker:
Remove instances of calling the Bluetooth utils.


SoundStream/res/values/strings.xml
```diff
@@ -63,4 +63,9 @@
     <string name="no_songs_in_playlist">There are no songs in your playlist! Quick, go add some from the music library!</string>
     <string name="disband">Disband</string>
     <string name="dialog_disband">Are you sure you want to disband? This will boot all connected users.</string>
+
+    <!-- Default values -->
+    <string name="default_mac">00:08:67:53:09</string>
+    <string name="default_name"></string>
```
ejohnson44: Is default name just supposed to be blank?

dgreenhalgh:
:shipit:

SoundStream/res/values/strings.xml
```diff
@@ -64,6 +64,10 @@
     <string name="disband">Disband</string>
     <string name="dialog_disband">Are you sure you want to disband? This will boot all connected users.</string>
 
+    <!-- default values -->
+    <string name="default_mac">00:08:67:53:09</string>
```
theJenix, line 68: :thumbsdown: to default MACs

SoundStream/src/com/lastcrusade/soundstream/service/MusicLibraryService.java
```diff
@@ -90,16 +90,25 @@ public MusicLibraryService getService() {
...
-        this.myMacAddress = BluetoothUtils.getLocalBluetoothMAC();
-        for (SongMetadata song : metadataList) {
-            song.setMacAddress(this.myMacAddress);
-        }
+        userListServiceLocator = new ServiceLocator<UserListService>(
+                this, UserListService.class, UserListService.UserListServiceBinder.class);
         
-        //update the library with the local songs
-        updateLibrary(metadataList, false);
+        myMacAddress = getResources().getString(R.string.default_mac);
```
theJenix, line 96: Keep this null until the user service is bound.

SoundStream/src/com/lastcrusade/soundstream/service/MusicLibraryService.java
```diff
@@ -319,6 +330,28 @@ private void sendSongData(String fromAddr, long songId) {
         }
     }
 
+    private String getMyMac(){
```
theJenix, line 333: This method appears to be unnecessary...it's only called when the service is bound, so we can call the service directly.

SoundStream/src/com/lastcrusade/soundstream/service/MusicLibraryService.java
reidbaker, line 96: So should I change https://github.com/reidbaker/SoundStream/blob/my_address/SoundStream/src/com/lastcrusade/soundstream/util/BluetoothUtils.java#L92 to return null as well.

SoundStream/res/values/strings.xml
reidbaker, line 68: Removed. Also removed the default name.

SoundStream/src/com/lastcrusade/soundstream/service/MusicLibraryService.java
reidbaker, line 333: Pulled to onBind method. Also onServiceBound should pass in the service.

reidbaker:
Made a trello card to figure out why removing the default variables causes connection to fail.

PR #137, opened 2013-04-12
updated version of app to publish

reidbaker:
This is what is on the play store.


ejohnson44:
You will probably want to update this once my color sync stuff gets pulled in so that image file isn't so huge - I know that was causing you problems last night

reidbaker:
Already pulled in the image things

ejohnson44:
Oh! Fantastic :monkey:

bryansills:
:thumbsup:

dgreenhalgh:
:shipit:

PR #139, opened 2013-04-15
Swipe removal fixes

ejohnson44:
Makes swipe removal cleaner

Does not fix Jesse's swipe problems - those will most likely require a major reworking of how we implement swipe. Trello card is made, will be looked at before official releases.


SoundStream/src/com/lastcrusade/soundstream/components/PlaylistFragment.java
```diff
@@ -304,6 +313,8 @@ public boolean onDoubleTap(MotionEvent e) {
                 return true;
             }
```
reidbaker, line 315: Did we need more whitespace here?

reidbaker:
:thumbsup:

dgreenhalgh:
:shipit:

PR #140, opened 2013-04-15
Google analytics

reidbaker:
Docs and code to add Google analytics to our code. 
After this is pulled in you will have to add the jar to the libs folder to compile


ejohnson44:
Looks good to me! :cherries:

dgreenhalgh:
:shipit:

PR #144, opened 2013-04-17
moved special characters to the end

ejohnson44:


SoundStream/src/com/lastcrusade/soundstream/util/AlphabeticalComparator.java
```diff
@@ -39,14 +41,36 @@ public int compare(SongMetadata songA, SongMetadata songB) {
         int albumComp =  songA.getAlbum().compareTo(songB.getAlbum());
         int titleComp = songA.getTitle().compareTo(songB.getTitle());
 
+        //not the most efficient way of doing things, but it works for now
```
bryansills: can you put a comment about what the code does here?

theJenix:
Doesn't String's built in compareTo function do this?  What does it do with special characters?

ejohnson44:
It puts them at the beginning, which is why the <unkown> artist stuff was coming up to the beginning of our lists, not at the end

theJenix:
I'd like to see if there's a better way to write this comparator...those string matches are really slow, and I don't like the multiple return statements.

ejohnson44:
The multiple returns prevent it from doing unnecessary comparisons - I tried to make it compare things as little as possible so that it wouldn't be too time consuming. 

I could rewrite it to convert to character arrays and manually compare things, but I think that is essentially what the string compare does anyways.

SoundStream/src/com/lastcrusade/soundstream/util/AlphabeticalComparator.java
```diff
@@ -39,14 +41,42 @@ public int compare(SongMetadata songA, SongMetadata songB) {
...
 
+        // not the most efficient way of doing things, but it works for now
+        // if the first character of songA's artist is a special character,
+        // and the first character of songB's artist is a normal letter, songA is ordered
+        // after songB
+        if(!songA.getArtist().substring(0, 1).matches("[a-zA-Z]") && songB.getArtist().substring(0, 1).matches("[a-zA-Z]")){
+            artistComp = 1;
+        }
+        //handles the opposite - A starts with a normal letter and B starts with a special character
+        else if(songA.getArtist().substring(0, 1).matches("[a-zA-Z]") && !songB.getArtist().substring(0, 1).matches("[a-zA-Z]")){
```
reidbaker: I dont mind this method of string comparisons for now. My only problem is it looks like repeated code.

reidbaker:
@theJenix What does Elizabeth need to do to get the ok from you?

SoundStream/src/com/lastcrusade/soundstream/util/AlphabeticalComparator.java
ejohnson44: It's repeated for each element of song, artist, and album, but I can figure out a way to abstract it out to a method

PR #145, opened 2013-04-17
Userlist bugs

reidbaker:
This fixes the userlist stuff in the menu fragment.

### DIFF
--- a/SoundStream/src/com/lastcrusade/soundstream/components/MenuFragment.java
+++ b/SoundStream/src/com/lastcrusade/soundstream/components/MenuFragment.java
@@ -108,9 +108,6 @@ public class MenuFragment extends SherlockFragment implements ITitleable {
             }
         });
         
-        //ListView userView = (ListView)v.findViewById(R.id.connected_users);
-        //userView.setAdapter(this.userAdapter);
-        
         userView = (LinearLayout)v.findViewById(R.id.connected_users);
         updateUserView();
         
--- a/SoundStream/src/com/lastcrusade/soundstream/components/MenuFragment.java
+++ b/SoundStream/src/com/lastcrusade/soundstream/components/MenuFragment.java
@@ -28,11 +28,10 @@ import android.view.View;
 import android.view.View.OnClickListener;
 import android.view.ViewGroup;
 import android.widget.Button;
-import android.widget.ListView;
+import android.widget.LinearLayout;
 
 import com.actionbarsherlock.app.SherlockFragment;
 import com.lastcrusade.soundstream.CoreActivity;
-import com.lastcrusade.soundstream.CustomApp;
 import com.lastcrusade.soundstream.R;
 import com.lastcrusade.soundstream.model.UserList;
 import com.lastcrusade.soundstream.service.ServiceLocator;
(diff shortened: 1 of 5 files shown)


SoundStream/src/com/lastcrusade/soundstream/components/MenuFragment.java
```diff
@@ -105,8 +108,11 @@ public void onClick(View v) {
             }
         });
         
-        ListView userView = (ListView)v.findViewById(R.id.connected_users);
-        userView.setAdapter(this.userAdapter);
+        //ListView userView = (ListView)v.findViewById(R.id.connected_users);
```
bryansills: remove commented code

SoundStream/src/com/lastcrusade/soundstream/components/NetworkFragment.java
```diff
@@ -122,8 +125,9 @@ public void onClick(View v) {
         disconnect = (Button)v.findViewById(R.id.disconnect_btn);
         disband = (Button)v.findViewById(R.id.disband_btn);
 
-        ListView users = (ListView) v.findViewById(R.id.connected_users);
-        users.setAdapter(this.adapter);
+        userView = (LinearLayout) v.findViewById(R.id.connected_users);
+        updateUserView();
+        //users.setAdapter(this.adapter);
```
bryansills, line 130: more commented code

bryansills:
other than the commented code looks good. is there an established format for tabbing with xml so that we don't have pull requests that look like huge changes, but only change the tabbing?

SoundStream/src/com/lastcrusade/soundstream/components/MenuFragment.java
```diff
@@ -178,4 +185,14 @@ private UserListService getUserListService(){
         }
         return userService;
     }
+    
+    private void updateUserView(){
```
theJenix, line 189: This feels wrong...given the circumstances, I'm not going to hold up the pull request but I'd like to know why we have to do this vs using a listview.

SoundStream/src/com/lastcrusade/soundstream/components/NetworkFragment.java
```diff
@@ -122,8 +125,9 @@ public void onClick(View v) {
         disconnect = (Button)v.findViewById(R.id.disconnect_btn);
         disband = (Button)v.findViewById(R.id.disband_btn);
 
-        ListView users = (ListView) v.findViewById(R.id.connected_users);
-        users.setAdapter(this.adapter);
+        userView = (LinearLayout) v.findViewById(R.id.connected_users);
```
theJenix, line 128: Dont like this...see above comment.

SoundStream/src/com/lastcrusade/soundstream/components/NetworkFragment.java
```diff
@@ -394,4 +400,14 @@ private UserListService getUserListService(){
         }
         return userService;
     }
+    
+    private void updateUserView(){
```
theJenix, line 404: Duplicate code (MenuFragment), and I repeat my objection.

ejohnson44, line 404: We do this in multiple places - In fact, anywhere that has the userlist service has two methods that are duplicated in every fragment that needs to get at the service and it's contents. I suppose we could work around it by making this a static method somewhere

theJenix:
After talking to @ejohnson44 my objections are retracted...somewhat.  I'm going to pull this in, but since we use the same strategy in both NetworkFragment and MenuFragment, please add a card create a reusable UserListView component.

SoundStream/src/com/lastcrusade/soundstream/components/NetworkFragment.java
theJenix, line 404: If we're going to stay with this strategy, we should define a UserListView that either extends View or LinearLayout, and exposes this functionality.  That way we can just reuse the component.

ejohnson44, line 404: That was my plan if I can't find another way to do this, but I didn't have the time to work with that to get this done for the demo today. I made a card in trello, so I will get to it soonish

PR #147, opened 2013-04-22
add page view hits in fragment

reidbaker:


dgreenhalgh:
:shipit:

dgreenhalgh:
Still :shipit:

PR #152, opened 2013-04-25
Click analytics

reidbaker:
This adds button presses to the analytic's we measure.

### DIFF
--- a/SoundStream/src/com/lastcrusade/soundstream/components/NetworkFragment.java
+++ b/SoundStream/src/com/lastcrusade/soundstream/components/NetworkFragment.java
@@ -56,6 +56,7 @@ import com.lastcrusade.soundstream.util.BroadcastRegistrar;
 import com.lastcrusade.soundstream.util.IBroadcastActionHandler;
 import com.lastcrusade.soundstream.util.ITitleable;
 import com.lastcrusade.soundstream.util.Toaster;
+import com.lastcrusade.soundstream.util.TrackerAPI;
 import com.lastcrusade.soundstream.util.Transitions;
 import com.lastcrusade.soundstream.util.UserListAdapter;
 
@@ -75,6 +76,8 @@ public class NetworkFragment extends SherlockFragment implements ITitleable {
     private ServiceLocator<ConnectionService> connectionServiceLocator;
     private ServiceLocator<UserListService>   userListServiceLocator;
 
+    private TrackerAPI tracker;
+
     @Override
     public void onCreate(Bundle savedInstanceState) {
         super.onCreate(savedInstanceState);
@@ -88,6 +91,7 @@ public class NetworkFragment extends SherlockFragment implements ITitleable {
             }
         });
 
+        tracker = new TrackerAPI((CoreActivity)getActivity());
         userAdapter = new UserListAdapter(getActivity(), new UserList(), false);
         userView = new LinearLayout(getActivity());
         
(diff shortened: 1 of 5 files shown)


SoundStream/src/com/lastcrusade/soundstream/components/NetworkFragment.java
```diff
@@ -118,6 +118,11 @@ public void onClick(View v) {
                 addMembersButton.setEnabled(false);
                 //TODO: add some kind of visual indicator while discovering...seconds until discovery is finished, number of clients found, etc
                 getConnectionService().findNewGuests();
+                ((CoreActivity)getActivity()).getTracker().sendEvent(
```
theJenix: We should abstract tracking events behind a service or API.  I don't like having these sendEvent calls directly in application code.

SoundStream/src/com/lastcrusade/soundstream/components/PlaybarFragment.java
```diff
@@ -99,13 +100,21 @@ public View onCreateView(LayoutInflater inflater, ViewGroup container,
             @Override
             public void onClick(View v) {
                 PlaylistService service = getPlaylistService();
+                long state; //Number for google analytics
                 if (service != null) {
                     if (service.isPlaying()) {
                         service.pause();
+                        state = 0; // 0 means we have paused
```
theJenix: We should expose isPaused through the playlist service (the IPlayer interface has an isPaused method, I believe).

SoundStream/src/com/lastcrusade/soundstream/components/NetworkFragment.java
reidbaker: I think these are fine until we get a better idea what we are tracking.

SoundStream/src/com/lastcrusade/soundstream/components/PlaybarFragment.java
reidbaker: Do you want me to change the code to use isPaused? I just added a state variable to send to google.

reidbaker:
The more I think about this I think we should pull it in. Does anyone have any objections?

SoundStream/src/com/lastcrusade/soundstream/components/NetworkFragment.java
theJenix: Sorry I should have replied earlier to this, because I really don't agree.  I think the easiest thing we could do here is pull these into simple class with descriptive method names, and instantiate one of those classes in onCreate.  That way, we can remove the magic strings from NetworkFragment, and it's still very tweakable as we find out more things we want to track.

SoundStream/src/com/lastcrusade/soundstream/components/PlaybarFragment.java
theJenix: Yea if we can reuse isPaused, i think that would be preferable.

theJenix:
I am building the simple Tracker API...I'll push it to you, Reid, in a separate branch and you can look at it.  After we're ok with that API, I'm good to pull this in.

theJenix:
My changes are in reid's tracker_api branch.

theJenix:
I merged the tracker_api changes into reid's branch, which augmented pretty much all of his changes.  Please review and pull in if we like it.

ejohnson44:
Looks good to me

dgreenhalgh:
:shipit:

PR #153, opened 2013-04-25
Crash on join

reidbaker:
Master is broken this fixes when you join on connect fragment and in the network fragment when you attempt to join another network

### DIFF
--- a/SoundStream/src/com/lastcrusade/soundstream/components/ConnectFragment.java
+++ b/SoundStream/src/com/lastcrusade/soundstream/components/ConnectFragment.java
@@ -111,7 +111,7 @@ public class ConnectFragment extends SherlockFragment implements ITitleable{
 
             @Override
             public void onClick(View v) {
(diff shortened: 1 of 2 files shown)


SoundStream/src/com/lastcrusade/soundstream/components/ConnectFragment.java
```diff
@@ -111,7 +111,7 @@ public void onClick(View v) {
 
             @Override
             public void onClick(View v) {
-                getConnectionService().broadcastSelfAsGuest(getActivity().getApplicationContext());
+                getConnectionService().broadcastSelfAsGuest(getActivity());
```
ejohnson44, line 114: I changed that in my other one - maybe it got pulled in before I pushed the fix. or the internet messed that bit up

ejohnson44:
Good to go.

bryansills:
:thumbsup:

PR #155, opened 2013-04-29
userlist_fixes

reidbaker:
I think that the userlist no longer updates. Here were my changes to fix that. 
I cant test it right now because I only have one device. Ill comment when I have tested.


theJenix:
This looks good to me.  :+1:

ejohnson44:
Looks good! Pulling it in.

PR #156, opened 2013-05-29
Join as <bluetooth name> in connect fragment

reidbaker:
I got this one done first to get back in the swing of coding.

### DIFF
--- a/SoundStream/res/layout/fragment_connect.xml
+++ b/SoundStream/res/layout/fragment_connect.xml
@@ -91,6 +91,7 @@
                 android:background="@color/accent_blue"/>
 
             <TextView 
+                android:id="@+id/join_network_id"
                 android:text="@string/join_network"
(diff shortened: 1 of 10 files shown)


reidbaker:
connect_as_working

dgreenhalgh:
Anyone think that this makes more sense with quotes around the name?

Meaning:

Join as "Nexus 4"

rather than 

Join as Nexus 4

I believe the default Bluetooth name is the device name, so the fact that the user would be identified by a Bluetooth name rather than a device name may be lost without quotes.

theJenix:
I think I agree with David...the quotes make sense, as we're specifying a specific string that will be used to identify this phone.

ejohnson44:
I agree with the others, but other than that this looks good to me

reidbaker:
This is now ready to pull in.

theJenix:
:beers:

PR #158, opened 2013-06-08
added a very simple pop up the first time you run the app

ejohnson44:


SoundStream/src/com/lastcrusade/soundstream/CoreActivity.java
```diff
@@ -96,6 +100,34 @@ public void onCreate(Bundle savedInstanceState){
...
+    /* (non-Javadoc)
+     * @see android.support.v4.app.FragmentActivity#onResume()
+     */
+    @Override
+    protected void onResume() {
+        // TODO Auto-generated method stub
+        super.onResume();
+        
+        SharedPreferences prefs = getSharedPreferences(
+                "com.thelastcrusade.soundstream", MODE_PRIVATE);
```
theJenix: I believe you can use getPackageName or some related method to get the package name defined in the manifest file.  I would recommend that instead of the literal string, or a comment saying why it should not match the application package.

theJenix:
Aside from my one comment, this looks :grapes:

SoundStream/res/values/strings.xml
```diff
@@ -76,4 +76,5 @@
     <string name="thanks_label">Special Thanks</string>
     <string name="thank_troy"> Troy Peace - A Super Sponsor, without whom this never could have happened</string>
 
+    <string name="welcome">Welcome to SoundStream!\n\nCreate a new network by pressing create, or join an existing network by pressing join and asking a friend to let you in.\n\nAdd music to your playlist through the Music page.\n\nIf there is no music there, check out the SoundStream wiki for information on how to add music to your phone.\n\nGo to the Network page to add people to your network.</string>
```
dgreenhalgh: Should create and join be capitalized? Nitpicking, otherwise this looks grapes.

dgreenhalgh:
:shipit:

PR #159, opened 2013-06-08
Network fragment redo

ejohnson44:
I'm putting this up as a pull request now because This is about the best I can do artwork-wise until I figure out a way to actually connect to the vm's at tech with any kind of reasonable performance. 

This is only a redo of the visuals of the page - I cannot touch the disconnect/disband functionality yet because I am only working with one phone. However, I have ordered another phone that I can use as a dev phone, so hopefully by later this week I will actually be able to do something useful.


theJenix:
Can you post a screenshot of this new page?

ejohnson44:
net

theJenix:
:hotsprings: :hotsprings: :hotsprings:

SoundStream/res/layout/fragment_network.xml
```diff
@@ -43,47 +37,185 @@
...
+			                />
+			            
+			        </RelativeLayout>			        
+			    </LinearLayout>
+		    
+		    </LinearLayout>
+		
+		        
+		    <!-- join different network 
+		    	TODO: rename this
```
reidbaker: Should this happen before it gets pulled in?

SoundStream/src/com/lastcrusade/soundstream/components/NetworkFragment.java
```diff
@@ -109,27 +109,57 @@ public View onCreateView(LayoutInflater inflater, ViewGroup container,
...
         
-        this.addMembersButton = (Button)v.findViewById(R.id.add_members);
+        this.addMembersButton = (LinearLayout)v.findViewById(R.id.add_members);
         this.addMembersButton.setOnClickListener(new OnClickListener() {
 
             @Override
             public void onClick(View v) {
                 addMembersButton.setEnabled(false);
-                //TODO: add some kind of visual indicator while discovering...seconds until discovery is finished, number of clients found, etc
+                //TODO: add some kind of visual indicator while discovering
```
reidbaker: I think this might be one of those bugs that would drive me crazy as a new user.

reidbaker:
This looks good to pull in once @ejohnson44 looks at my comments.

SoundStream/src/com/lastcrusade/soundstream/components/NetworkFragment.java
ejohnson44: @theJenix Do we have a way to get this information from the bluetooth stuff we have now? If we do I can think of some pretty easy ways to add this is, but if not it might be a good bit harde

SoundStream/res/layout/fragment_network.xml
ejohnson44: Haha, yeah it probably should. I will fix this after work

SoundStream/src/com/lastcrusade/soundstream/components/NetworkFragment.java
theJenix: We get callbacks when discovery starts, stops, and when we discover a device...that's currently down in the bowels of ConnectionService, but we can send an intent for those events.

ejohnson44:
Screenshot of the new search progress
searching

dgreenhalgh:
:+1:

theJenix:
DOUBLE :+1:

PR #160, opened 2013-06-09
Fixes for various bugs with audio ducking/focus, and some misc NPEs

theJenix:
Fixed problem where audio was being unpaused when a user pauses the song..., and an audio event from another app happens.  In the process, I found a really weird
bug with Android Audio focus...if two or more other apps request audiofocus gain
can duck at or around the same time, audio focus may not always be returned to
SoundStream.  This is fixed by using a timer task, scheduled 2 seconds out from
when we got the can_duck notification, and will re-grab audio focus and unduck
the volume manually

modified external control client to not create the client immediately (but not
register it).  This is useful because it means the client object is never null,
which fixes a crash caused by pressing play very quickly after adding your first
song.

added test for null in SingleFileAudioPlayer to fix the same kind of bug.

### DIFF
--- a/SoundStream/src/com/lastcrusade/soundstream/audio/AudioPlayerWithEvents.java
+++ b/SoundStream/src/com/lastcrusade/soundstream/audio/AudioPlayerWithEvents.java
@@ -37,7 +37,7 @@ public class AudioPlayerWithEvents implements IPlayer {
 
     private static final String TAG = AudioPlayerWithEvents.class.getSimpleName();
 
-    private static final int DUCK_TIMER_DELAY = 2000;
+    private static final int DUCK_TIMER_DELAY_MS = 2000;
 
     private IPlayer player;
     private Context context;
@@ -137,7 +137,7 @@ public class AudioPlayerWithEvents implements IPlayer {
     }
 
     /**
-     * Handle loss of audio focus.  In the both permanent case,
+     * Handle loss of audio focus.  In the the case of permanent loss,
      * we want to unregister our external control client and pause the music.
      * 
      * In all cases, pause the music and indicate that the music can be resumed
@@ -229,7 +229,7 @@ public class AudioPlayerWithEvents implements IPlayer {
             @Override
             public void run() {
                 if (LogUtil.isLogAvailable()) {
-                    Log.d(TAG, "Duck timer executed...something's wrong.  Resetting audio focus and unducking volumne..");
+                    Log.d(TAG, "Duck timer executed...something's wrong.  Resetting audio focus and unducking volume..");
                 }
                 //in this case, something is stuck...rerequest audio focus
                 //NOTE: this is important, because if we didn't receive the AUDIOFOCUS_GAIN
@@ -240,7 +240,7 @@ public class AudioPlayerWithEvents implements IPlayer {
                 unduck();
             }
             
-        }, DUCK_TIMER_DELAY);
+        }, DUCK_TIMER_DELAY_MS);
     }
     /**
      * Duck the volume of the player, if the player supports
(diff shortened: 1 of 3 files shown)


SoundStream/src/com/lastcrusade/soundstream/audio/AudioPlayerWithEvents.java
```diff
@@ -34,13 +37,19 @@
 
     private static final String TAG = AudioPlayerWithEvents.class.getSimpleName();
 
+    private static final int DUCK_TIMER_DELAY = 2000;
```
reidbaker: units?

SoundStream/src/com/lastcrusade/soundstream/audio/AudioPlayerWithEvents.java
```diff
@@ -107,10 +104,45 @@ public void onAudioFocusChange(int focusChange) {
...
+            if (LogUtil.isLogAvailable()) {
+                Log.w(TAG, String.format("Unhandled focus change event %d", focusChange));
+            }
+
+        }
+    }
+
     /**
-     * Handle loss of audio focus.  In both permanent and temporary cases,
+     * Handle loss of audio focus.  In the both permanent case,
```
reidbaker: In both permanent case?

SoundStream/src/com/lastcrusade/soundstream/audio/AudioPlayerWithEvents.java
```diff
@@ -120,10 +152,13 @@ private void handleAudioFocusLoss(boolean permanent) {
         if (LogUtil.isLogAvailable()) {
             Log.d(TAG, String.format("Audio focus lost%s", permanent ? "" : " (transient)"));
         }
-        //unregister the external control client when focus is lost, as per best practice
-        unregisterExternalControlClient();
-        //NOTE: this calls the the pause method directly, because we want instant action.
-        pause();
+        //we've permanently lost audio focus...unregister thyself
```
reidbaker, line 155: unregister thyself and repent :fire: 
No comment here

SoundStream/src/com/lastcrusade/soundstream/audio/AudioPlayerWithEvents.java
```diff
@@ -150,16 +190,59 @@ private void handleAudioFocusReturned(boolean permanent) {
...
+            cancelDuckTimer();
+        }
+
+        this.duckTimer = new Timer();
+        this.duckTimer.schedule(new TimerTask() {
+
+            @Override
+            public void run() {
+                if (LogUtil.isLogAvailable()) {
+                    Log.d(TAG, "Duck timer executed...something's wrong.  Resetting audio focus and unducking volumne..");
```
reidbaker: unducking _volume_

SoundStream/src/com/lastcrusade/soundstream/audio/AudioPlayerWithEvents.java
```diff
@@ -192,19 +275,42 @@ public boolean isPlaying() {
...
     @Override
     public void pause() {
+        canAudioFocusResume = false;
+        doPause();
+    }
+    
+    /**
+     * Private helper method that will pause the audio, regardless of who calls it.
+     * 
+     * Callers can then decide whehter its appropriate to pause on their own terms.
```
reidbaker: whehter?

theJenix:
Well those typos were embarrassing.  So is the typo in my commit message.  Holy :sheep:

reidbaker:
I assumed they were in there to check if we read your code

dgreenhalgh:
:shipit:

PR #161, opened 2013-06-09
back button change 

ejohnson44:
changed the back button to back out of the app when in the connection fragment or when the menu is open, and to open the menu otherwise

### DIFF
--- a/SoundStream/src/com/lastcrusade/soundstream/CoreActivity.java
+++ b/SoundStream/src/com/lastcrusade/soundstream/CoreActivity.java
@@ -23,6 +23,8 @@ import android.content.Context;
 import android.content.Intent;
 import android.os.Bundle;
 import android.support.v4.app.Fragment;
+import android.util.Log;
+import android.view.KeyEvent;
 
 import com.actionbarsherlock.view.MenuItem;
 import com.google.analytics.tracking.android.EasyTracker;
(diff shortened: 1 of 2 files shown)


SoundStream/src/com/lastcrusade/soundstream/CoreActivity.java
```diff
@@ -94,9 +98,43 @@ public void onCreate(Bundle savedInstanceState){
...
+    }
+    
+    @Override
+    public void onBackPressed() {
+        //if we are in the connect fragment or the menu is open, 
+        //back out of the app by pulling up the home screen
+        if(getTitle().equals(getString(R.string.select)) || getSlidingMenu().isMenuShowing()){
+            Intent startMain = new Intent(Intent.ACTION_MAIN);
+            startMain.addCategory(Intent.CATEGORY_HOME);
+            startMain.setFlags(Intent.FLAG_ACTIVITY_NEW_TASK);
```
theJenix, line 165: I think we may want FLAG_ACTIVITY_CLEAR_TASK as well...https://developer.android.com/reference/android/content/Intent.html#FLAG_ACTIVITY_CLEAR_TASK

ejohnson44, line 165: I just tried it and that made it ridiculously slow

theJenix, line 165: I don't actually remember what I was thinking with my previous comment.  Yea, that makes sense.  Just ignore me :)

theJenix:
:thumbsup:

dgreenhalgh:
:shipit:

PR #163, opened 2013-06-09
fixed network user layout in menu

theJenix:
added transparent divider to linearlayout for network users, and changed layout to vertical


reidbaker:
push it like its :fireworks:

ejohnson44:
looks good

PR #209, opened 2013-07-14
Skip fix

### LINKED ISSUE #175: Skip when paused automatically plays the song it skips to
If this is the expected behavior, then this defect is with the test plan.

To reproduce:
Test 12a in test plan.

ejohnson44:
skip while paused now remains paused

fixes #175


dgreenhalgh:
:+1:

PR #215, opened 2013-07-15
Disband cleanup

### LINKED ISSUE #188: Disband does not stop the currently playing song
Observed when executing Test 25, when a song is playing.  The current song plays to completion, or until the user skips the song.  It does not show up in the new playlist, and does not repeat.

theJenix:
Added code to clean up the player when the playlist is cleared, after a network is disbanded.

fixes #188

### DIFF
--- a/SoundStream/src/com/lastcrusade/soundstream/audio/RemoteAudioPlayer.java
+++ b/SoundStream/src/com/lastcrusade/soundstream/audio/RemoteAudioPlayer.java
@@ -51,11 +51,6 @@ public class RemoteAudioPlayer implements IPlayer {
     }
 
     @Override
-    public void cancel() {
-        //NO OP
-    }
-
-    @Override
     public boolean isPaused() {
         return this.paused;
     }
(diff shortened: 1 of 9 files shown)


SoundStream/src/com/lastcrusade/soundstream/audio/RemoteAudioPlayer.java
```diff
@@ -106,23 +106,31 @@ public void skip() {
...
+            public void onReceiveAction(Context context, Intent intent) {
+                paused = true;
+                playing = false;
+            }
+        })
+        .addLocalAction(PlaylistService.ACTION_PLAYING_AUDIO, new IBroadcastActionHandler() {
+            
+            @Override
+            public void onReceiveAction(Context context, Intent intent) {
+                paused = false;
```
ejohnson44, line 130: Is there a reason that we need to keep track of both playing and paused? Are they not mutually exclusive?

reidbaker, line 130: I think you can also be stopped

theJenix, line 130: Correct...the regular player lets you test isPlaying and isPaused because it can also be stopped, which implies that the resources for the player are freed.  I needed to mimic this in the RemoteAudioPlayer class.

PR #217, opened 2013-07-21
added seaching feedback to the join button

### LINKED ISSUE #193: Join Network gives no indication it is working
When joining a network, there is no indication or feedback that it is searching, or that it has found any results.  The UI just hangs.  There should at least be a spinner or some other progress indicator.

ejohnson44:
fixes #193


reidbaker:
looks gr:eight: to me

dgreenhalgh:
:shipit:

Issue #179, opened 2013-07-07
Cannot remove currently playing item on guest


theJenix:
When you try to remove the currently playing item, the removed item reappears and pops up a toast saying "The playlist is empty.

Observed when performing Test 13b.

dgreenhalgh:
Verify on #178 fix

theJenix:
With pull request #214, I've seen this issue largely go away except when attempting to remove the currently playing item.  To reproduce:

1) establish a network, and add 3 items that live on the host to the playlist (you can add items from anywhere, but adding items that live on the host will be quicker).
2) press play, and skip to the last song.
3) from the guest, swipe to remove the currently playing song.

What I see is it skips to the top of the playlist, but the song that was removed comes back.

reidbaker:
What I see as of now is this only happens for songs on the host device. If the song removed is the guest's song then the song flashes but is removed.

theJenix:
This was fixed in #220, and then the fix was changed slightly in #222.  This issue should get closed along with #222

PR #220, opened 2013-07-21
fix for #179 removing currently playing song on guest doesnt work

reidbaker:
fix for #179 removing currently playing song on guest doesnt work


theJenix:
f-ing :thumbsup:

PR #225, opened 2013-07-28
fixed rotation menu bug

ejohnson44:
fixed the bug where rotation locked you our of the menu via the home button


reidbaker:
Great :shipit:

dgreenhalgh:
:ship:

PR #255, opened 2013-08-05
State when paused

### LINKED ISSUE #234: Guest state after skip while paused
When a song is skipped while paused, the guest playbar and now playing icon do not have the correct state.

The guest playbar will continue to show the last song that was played, and the now playing icon disappears.

To replicate:
- connect 2 phones
- add some songs to the library
- press play
- press pause
- press skip

### LINKED ISSUE #244: Skipping song when paused does not update the current song in the player
Steps to reproduce:

1) Connect two phones and add three songs to the playlist from any device.
2) Play the first song and pause before it ends (from any device).
3) Skip to the third song (from any device).
4) Press play (from any device).

Expected: The third song should play.

Actual: The first song resumes playing.

This resolves if either host or guest skips while the song is playing.

### LINKED ISSUE #248: Removing currently playing song when paused will not change the song when played
Observed during test 19a.  Pressing skip when playing fixes the state and resolves the issue.

theJenix:
Fixed issues with skipping/removing when paused by ensuring that any "set next song call" effectively will cause the player to play from the beginning of whatever song is set.

Fixed issue with synchronization between host and guest by sending playStatusMessage when host sets a new song.

Fixes #234
Fixes #244
Fixes #248

### DIFF
--- a/SoundStream/src/com/lastcrusade/soundstream/service/PlaylistService.java
+++ b/SoundStream/src/com/lastcrusade/soundstream/service/PlaylistService.java
@@ -21,9 +21,7 @@ package com.lastcrusade.soundstream.service;
 
 import java.util.ArrayList;
 import java.util.Collections;
-import java.util.HashSet;
 import java.util.List;
-import java.util.Set;
 
 import android.app.Service;
 import android.content.Context;
@@ -34,7 +32,6 @@ import android.util.Log;
 
 import com.lastcrusade.soundstream.R;
 import com.lastcrusade.soundstream.audio.AudioPlayerWithEvents;
-import com.lastcrusade.soundstream.audio.IPlayer;
 import com.lastcrusade.soundstream.audio.RemoteAudioPlayer;
 import com.lastcrusade.soundstream.audio.SingleFileAudioPlayer;
 import com.lastcrusade.soundstream.manager.PlaylistDataManager;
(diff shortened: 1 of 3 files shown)


SoundStream/src/com/lastcrusade/soundstream/service/PlaylistService.java
```diff
@@ -580,6 +584,11 @@ public void pause() {
     }
 
     public void skip() {
+    	//TODO: I'm not sure I like this...the player ends up sending the same intent
+    	// for when a song is "finished", which is handled above, which calls setNextSong
+    	//...this presents an unobvious execution path, and conflates the acts of
+    	// skipping to the next song, and finishing playing the next song, and has
+    	// led to some unobvious bugs -- Jesse 08/04/13
```
dgreenhalgh, line 591: You say "has led to unobvious bugs" - could this be a problem child for the future?

dgreenhalgh:
I like this. @reidbaker, what do you think?

SoundStream/src/com/lastcrusade/soundstream/service/PlaylistService.java
theJenix, line 591: This round about chain of events kind of caused the current bug being fixed (in some ways), and made it really hard to construct an easy fix (hence the amount of change).  I think it's good for now, but I'll add a "refactor" issue to take another look at this whole thing.

reidbaker:
yeah I am pulling this in

SoundStream/src/com/lastcrusade/soundstream/service/PlaylistService.java
dgreenhalgh, line 591: Yeah, this may be something we want to look into between beta and 1.0

PR #261, opened 2013-08-08
Monster refactor

reidbaker:
Change our package from com.lastcrusade to com.thelastcrusade and made the version number 1.


dgreenhalgh:
Thanks a lot man. To fill you guys in, we ran in a problem uploading the APK due to an issue with our version numbering and the fact that our package names conflicted with those of SoundStream Alpha. This big naming refactor is the most time-efficient way to solve this issue. 

Let's please try and pull this in ASAP so that we can push to the Play Store

theJenix:
Looks good to me, and builds on my machine.  Two enthusiastic :thumbsup:

dgreenhalgh:
I'll be running through some quick tests of this version of the code tonight before this goes out

dgreenhalgh:
Lady and gentlemen, we have a published beta.

PR #304, opened 2013-08-25
added some clarificaiton instructions to the connection process

ejohnson44:
Instead of simple "searching", the button (join and add members to network) now also displays some instructions when pressed

### DIFF
--- a/SoundStream/res/layout/fragment_connect.xml
+++ b/SoundStream/res/layout/fragment_connect.xml
@@ -87,9 +87,16 @@
 	        android:layout_centerInParent="true"
 	        android:background="@color/searching_gray"
 	        android:padding="10dp"
-	        android:visibility="invisible" >
-	
-	        <ProgressBar
+	        android:alpha=".95"
+	        android:visibility="invisible"
+	         >
+			<RelativeLayout 
+			    android:id="@+id/searching_text"
+			    android:layout_centerHorizontal="true"
+			    android:layout_width="wrap_content"
+	        	android:layout_height="wrap_content"
+	        	>
+			    <ProgressBar
 	            android:id="@+id/progress"
 	            style="@style/Widget.Sherlock.Light.ProgressBar"
 	            android:layout_width="wrap_content"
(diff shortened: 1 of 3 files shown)


dgreenhalgh:
:+1: Good to go if someone else takes a look at this

theJenix:
Can I see a screenshot of this?

ejohnson44:
I actually can't get one. My new phone is doing the crash on startup that seems to be related to google analytics things. I tried on the most recent master that has the most recent code that should have fixed that but it doesn't do anything different

dgreenhalgh:
Hmm this sounds high priority. What kind of phone did you get?

ejohnson44:
Motorola Droid Razr - running 4.2.1

theJenix:
I think this is a showstopper, unfortunately.  Elizabeth, since you can reproduce it, do you have time to research the issue today?   If not, do you have time to run the app a couple times in various different states (e.g. phone just started, phone running for a bit, no other apps running, etc) and post the exceptions?

ejohnson44:
I don't have time to do much work on it, but I can post the exceptions. I can also try it on Mark's phone (he has the same kind I do, and I'm pretty sure that at one point in time his phone handled the app just fine)

ejohnson44:
Ok, so actually I think it had to do with the fact that I had no music on my phone. I tried adding music in a couple of different places and had no luck, but I tried one more place this morning and it worked.

The error I was getting was a null pointer exception from the cursor in the media store wrapper when 

https://www.evernote.com/shard/s239/sh/a6b69f1e-6a19-43ea-b112-1203bf56634c/e750d3dc7cc00cbe59f2d63180c418f1?noteKey=e750d3dc7cc00cbe59f2d63180c418f1&noteGuid=a6b69f1e-6a19-43ea-b112-1203bf56634c

Jesse, do you know how to fix this?

ejohnson44:
add
join

Also this phone is going to drive me crazy. I just realized that this happens ONLY when my phone is connected to the computer as it tries to start up because when it is connected, it won't let you access the files you have in memory. So this shouldn't be an issue for any users

dgreenhalgh:
@theJenix, if this NPE is a quick fix, we may want to try and get it into SoundStream 0.8.1 tonight

dgreenhalgh:
I moved this crash into it's own issue, #312

dgreenhalgh:
That bug is unrelated to this pull request, so I'm going to go ahead and pull this in. :+1:

PR #305, opened 2013-08-25
Instructions are no longer one single page of text

ejohnson44:
broke up the single instructions dialog into multiple consecutive dialogs and allowed the user to access them again via the about page.

In the future I want to change this to the nice multi-page dialogs that some apps use (does anyone know how they do this? I couldn't find much about it when I was looking) with colors and icons and things

### DIFF
--- a/SoundStream/res/values/strings.xml
+++ b/SoundStream/res/values/strings.xml
@@ -90,5 +90,10 @@
     <string name="credit_label">Credit</string>
     <string name="brought_to_you_by">SoundStream is built on top of the following fine open source projects:</string>
 
-    <string name="welcome">Welcome to SoundStream!\n\nCreate a new network by pressing Create, or join an existing network by pressing Join and asking a friend to let you in.\n\nAdd music to your playlist through the Music page.\n\nIf there is no music there, check out the SoundStream wiki for information on how to add music to your phone.\n\nGo to the Network page to add people to your network.</string>
+    <string name="welcome">Welcome to SoundStream!\n\nFor some basic instructions and hints, press Next.</string>
+	<string name="welcome_connection">Create a new network by pressing Create, or join an existing network by pressing Join and asking a friend to let you in.</string>
+	<string name="welcome_music">Add music to your playlist through the Music page.\n\nIf there is no music there, check out the SoundStream wiki for information on how to add music to your phone.</string>
+	<string name="welcome_network">To view information about the network, go the the network page.\n\n From there you can add members to the network, disconnect, or switch networks. </string>
+	<string name="instructions_label">Instructions</string>
+	<string name="instructions">View the startup instructions again</string>
 </resources>
--- a/SoundStream/res/values/strings.xml
+++ b/SoundStream/res/values/strings.xml
@@ -92,8 +92,8 @@
 
     <string name="welcome">Welcome to SoundStream!\n\nFor some basic instructions and hints, press Next.</string>
 	<string name="welcome_connection">Create a new network by pressing Create, or join an existing network by pressing Join and asking a friend to let you in.</string>
-	<string name="welcome_music">Add music to your playlist through the Music page.\n\nIf there is no music there, check out the SoundStream wiki for information on how to add music to your phone.</string>
-	<string name="welcome_network">To view information about the network, go the the network page.\n\n From there you can add members to the network, disconnect, or switch networks. </string>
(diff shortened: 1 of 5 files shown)


dgreenhalgh:
Let's walkthrough these changes and #304 tonight - we can talk about desired behaviors then

SoundStream/src/com/thelastcrusade/soundstream/components/InstructionsDialog.java
```diff
@@ -0,0 +1,68 @@
...
+    public InstructionsDialog(Activity activity){
+        this.activity = activity;
+    }
+    public void show(){
+        showInstructions(0);
+    }
+    
+    private void showInstructions(final int messageIndex){
+        String positiveString = messageIndex < MESSAGEIDS.length-1?  "Next" : "Done";
+        String negativeString = messageIndex > 0 ?  "Back" : "Skip";
```
reidbaker: I see some magic numbers from here down (mostly - one's and zero). Where do they come from?

dgreenhalgh: Do ones and zeroes count as magic numbers in this case when they're used for indexing?

SoundStream/res/values/strings.xml
```diff
@@ -90,5 +90,10 @@
     <string name="credit_label">Credit</string>
     <string name="brought_to_you_by">SoundStream is built on top of the following fine open source projects:</string>
 
-    <string name="welcome">Welcome to SoundStream!\n\nCreate a new network by pressing Create, or join an existing network by pressing Join and asking a friend to let you in.\n\nAdd music to your playlist through the Music page.\n\nIf there is no music there, check out the SoundStream wiki for information on how to add music to your phone.\n\nGo to the Network page to add people to your network.</string>
+    <string name="welcome">Welcome to SoundStream!\n\nFor some basic instructions and hints, press Next.</string>
+	<string name="welcome_connection">Create a new network by pressing Create, or join an existing network by pressing Join and asking a friend to let you in.</string>
+	<string name="welcome_music">Add music to your playlist through the Music page.\n\nIf there is no music there, check out the SoundStream wiki for information on how to add music to your phone.</string>
```
dgreenhalgh: Should maybe be the Music Library page

SoundStream/res/values/strings.xml
```diff
@@ -90,5 +90,10 @@
...
+	<string name="welcome_network">To view information about the network, go the the network page.\n\n From there you can add members to the network, disconnect, or switch networks. </string>
```
dgreenhalgh: We may want to capitalize Network page for consistency

SoundStream/src/com/thelastcrusade/soundstream/components/InstructionsDialog.java
theJenix: Yea they do. We should give them a name so we know what index we're using. 

David Greenhalgh notifications@github.com wrote:

> - */
>   +public class InstructionsDialog {
> -    private Activity activity;
> -    private static final int[] MESSAGEIDS = {R.string.welcome, R.string.welcome_connection, 
> -            R.string.welcome_music, R.string.welcome_network };
> -   
> -    public InstructionsDialog(Activity activity){
> -        this.activity = activity;
> -    }
> -    public void show(){
> -        showInstructions(0);
> -    }
> -   
> -    private void showInstructions(final int messageIndex){
> -        String positiveString = messageIndex < MESSAGEIDS.length-1?  "Next" : "Done";
> -        String negativeString = messageIndex > 0 ?  "Back" : "Skip";

Do ones and zeroes count as magic numbers in this case when they're used for indexing?

---

Reply to this email directly or view it on GitHub:
https://github.com/TheLastCrusade/SoundStream/pull/305/files#r6032122

dgreenhalgh: Alright

theJenix: Well...so, looking at this again, I might even go so far as to define a method that tests "isAtEnd" or "isAtMiddle" or something.  In that case, we can use 0 and 1s in the test, and it's likely more descriptive than calling 1 "OFF OFFSET" or something like that.

dgreenhalgh:
With these changes, I think this is good to go :+1:

theJenix:
This looks gr:eight:.  I'm merging.

PR #307, opened 2013-08-25
Perform lastcrusade -> thelastCrusade refactor on tests

taywrobel:
Do the master refactoring to renaming packages to thelastcrusade
for the test project as well.

Also updates the test project's Android Manifest to reflect the
changes and updates the gitignore to ignore test libraries and
the third party directory in the main git directory.

Tested change.


reidbaker:
This block us from being able to run our tests and does not affect app functionality so I am going to pull this in. If we have an issue with that we can talk about it during the meeting planned tonight.

PR #308, opened 2013-08-26
implemented changes to ask the user before enabling bluetooth.

### LINKED ISSUE #293: Need a warning that we're turning on bluetooth
A couple people have commented that they were surprised that our app automatically turns on bluetooth.  Not necessarily in a "oh it shouldn't do that" way, but we probably should put up a small warning that "this app will enable your bluetooth radio."

### LINKED ISSUE #294: Bluetooth enabled when app starts
A couple people have commented about starting bluetooth when the app starts.  We don't actually need bluetooth until we attempt to scan for a guest or make ourselves findable by a host, so we should not enable bluetooth until then.

I made this an issue because I believe this isn't working as intended and is causing people confusion.

### LINKED ISSUE #298: Bluetooth enabled at random after SoundStream is installed
Some users have reported an issue where, after installing SoundStream, their bluetooth adapter will randomly enable itself.

theJenix:
removed auto enable of bluetooth from ConnectionService onCreate

added wrapper class to ask the user, enable bluetooth, and then run supplied
code

implemented use of wrapper on "Join" button (connectfragment), "Add users"
button, and "Switch networks" button (networkfragment)

Fixes #298 
Fixes #294 
Fixes #293

### DIFF
--- a/SoundStream/src/com/thelastcrusade/soundstream/components/WithBluetoothEnabled.java
+++ b/SoundStream/src/com/thelastcrusade/soundstream/components/WithBluetoothEnabled.java
@@ -1,0 +1,96 @@
+/*
+ * Copyright 2013 The Last Crusade ContactLastCrusade@gmail.com
+ * 
+ * This file is part of SoundStream.
+ * 
+ * SoundStream is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU General Public License as published by
+ * the Free Software Foundation, either version 3 of the License, or
+ * (at your option) any later version.
+ * 
+ * SoundStream is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU General Public License for more details.
+ * 
+ * You should have received a copy of the GNU General Public License
+ * along with SoundStream.  If not, see <http://www.gnu.org/licenses/>.
+ */
+package com.thelastcrusade.soundstream.components;
+
+import android.app.AlertDialog;
+import android.app.AlertDialog.Builder;
+import android.content.Context;
+import android.content.DialogInterface;
+import android.os.Handler;
+import android.util.Log;
+
+import com.thelastcrusade.soundstream.R;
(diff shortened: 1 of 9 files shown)


SoundStream/src/com/thelastcrusade/soundstream/components/WithBluetoothEnabled.java
```diff
@@ -0,0 +1,96 @@
...
+                        @Override
+                        public void onClick(DialogInterface dialog, int which) {
+                            if (connectionService.enableNetwork()) {
+                                handler.post(runnable);
+                            } else {
+                                Log.wtf(TAG, "Error enabling bluetooth");
+                            }
+                        }
+                    })
+                   .setNegativeButton(android.R.string.cancel, new DialogInterface.OnClickListener() {
```
reidbaker, line 86: I dont think we have to set an onclick listener if we dont want to do anything.

reidbaker:
one small comment but I still think this is good to :punch:  in

SoundStream/src/com/thelastcrusade/soundstream/components/WithBluetoothEnabled.java
theJenix, line 86: I think if we don't, it wont display a cancel button (i'll verify this today).  If that's the case, do we prefer no cancel button, or a NOOP listener?

reidbaker, line 86: I would prefer a noop listener to not having a cancel button

dgreenhalgh:
Once Reid's comment is cleared up, let's :shipit:

theJenix:
Hmm...it appears my latest commit cleared out comments about the cancel button...that's unfortunate.  Anyway, you must have a setNegativeButton call to get the cancel button, so I left in the no-op listener and added a comment explaining it.

theJenix:
Are there any other comments on this?  Is this ready to go?

dgreenhalgh:
If @reidbaker's happy, let's :shipit:

taywrobel:
I had to look up android threading (Handler, really, are they that averse to normal conventions -.-), but after that I think this looks good.  Pulling in so we can get some testing in before the next patch release.

PR #309, opened 2013-08-27
Limit the number of concurrent toast messages 

reidbaker:
:ship: + :ocean: = :collision:


dgreenhalgh:
:+1: @theJenix, how do you feel about this one?

SoundStream/src/com/thelastcrusade/soundstream/util/Toaster.java
```diff
@@ -29,18 +32,40 @@
...
+        }
+        return true;
     }
 
-    private static void toastWorker(Context context, String s) {
+    public static synchronized void toastWorker(Context context, String s) {
+        if (isRecent(s)) {
+            return;
+        }
+        lastShown.clear(); // Keep our map from growing in size
```
theJenix, line 56: I don't think I like this call.  If you had 4 toast messages you're throttling, this will clear all of their last shown times...does that cause issues?

theJenix:
So, unfortunately it doesn't look like I can push my changes directly to Reid's master branch (which is referenced by this pull request.)  I will issue a pull request from a branch on my repo.

theJenix:
See pull request #311

Issue #285, opened 2013-08-12
Text search or filter music library


theJenix:
Allow the user to filter artist, title or album name with free text input.

theJenix:
:+1: Important for phones with large music libraries.

dgreenhalgh:
:+1:

PR #320, opened 2013-10-06
Search for music v1

reidbaker:


reidbaker:
fixes issue #285

SoundStream/res/xml/searchable.xml
```diff
@@ -0,0 +1,5 @@
+<?xml version="1.0" encoding="utf-8"?>
+<searchable xmlns:android="http://schemas.android.com/apk/res/android"
+    android:label="@string/app_name"
+    android:hint="@string/search_hint" >
+</searchable>
```
ejohnson44: New Line?

SoundStream/src/com/thelastcrusade/soundstream/components/MusicLibraryFragment.java
```diff
@@ -60,6 +62,8 @@
     private ServiceLocator<MusicLibraryService> musicLibraryServiceLocator;
     
     private MusicAdapter mMusicAdapter;
+    
+    private volatile String mQuery;
```
ejohnson44, line 66: what does setting this to volatile do? (aka why are we doing this)

ejohnson44:
Overall everything looks pretty good to me considering that I don't know all that much about search things. The only thing I might say is that it would be nice if you added in some comments since I don't think many of us are familiar with the search activity concept (but I could be wrong)

dgreenhalgh:
Yep, I'm in the same boat as @ejohnson44 here. Everything looks good, but it may help to have some context at more of a macro level

dgreenhalgh:
@reidbaker walked us through this code at the Nov. 17 meeting. After the newline is fixed, I say :+1:

PR #327, opened 2013-12-10
removing actionbar sherlock from the project and replacing it with appcompat

reidbaker:
DO NOT PULL THIS IN.
I am putting it up so that yall can see what the first round of changes looks like.

replace abs with appcompat, compiles, doesnt run because search is missing styling, docs not updated

### DIFF
--- a/SoundStream/res/values/colors.xml
+++ b/SoundStream/res/values/colors.xml
@@ -33,6 +33,11 @@
     <color name="playbar">#DEDEDE</color>
     <color name="accent_blue">#33008AA6</color>
     <color name="medium_gray">#EEEEEE</color>
-    <color name="link_blue">#5858F8</color>
+    <color name="link_blue">#5858F8</color>  
+    <!-- Refactor colors -->      
+    <color name="holo_light">#fff3f3f3</color>
+    <color name="primary_text_holo_light">#80ffffff</color>
+    <color name="bright_foreground_disabled_holo_light">#ffb2b2b2</color>
+    <color name="holo_blue_light">#ff33b5e5</color>
     
 </resources>
--- a/SoundStream/src/com/thelastcrusade/soundstream/CoreActivity.java
+++ b/SoundStream/src/com/thelastcrusade/soundstream/CoreActivity.java
@@ -23,26 +23,25 @@ import java.lang.ref.WeakReference;
 import java.util.ArrayList;
 import java.util.List;
 
-import android.app.AlertDialog;
+import android.app.ActionBar;
 import android.app.SearchManager;
 import android.content.ComponentName;
 import android.content.Context;
-import android.content.DialogInterface;
 import android.content.Intent;
 import android.content.SharedPreferences;
 import android.os.Bundle;
 import android.support.v4.app.Fragment;
 import android.util.Log;
 import android.view.KeyEvent;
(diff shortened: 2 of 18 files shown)


dgreenhalgh:
Thanks for putting this up, @reidbaker

dgreenhalgh:
What's the status of this work, @reidbaker?

SoundStream/res/values/colors.xml
```diff
@@ -33,6 +33,11 @@
     <color name="playbar">#DEDEDE</color>
     <color name="accent_blue">#33008AA6</color>
     <color name="medium_gray">#EEEEEE</color>
-    <color name="link_blue">#5858F8</color>
+    <color name="link_blue">#5858F8</color>  
+    <!-- Refactor colors -->
```
theJenix, line 37: Is this so that we maintain the same app appearance after removing ABS?  If so, is there something more standard we can switch to in AppCompat, at the expense of changing the app a bit?

SoundStream/src/com/thelastcrusade/soundstream/CoreActivity.java
```diff
@@ -195,7 +192,8 @@ public boolean onCreateOptionsMenu(Menu menu) {
     }
     
     private boolean isConnectActive(){
-        return getSupportFragmentManager().findFragmentByTag(Transitions.CURRENT_CONTENT) instanceof ConnectFragment;
+        return true;
```
theJenix: I know this code is still in development, but I wanted to tag this so we don't overlook it.  This needs to be corrected, or commented.

SoundStream/src/com/thelastcrusade/soundstream/components/PlaylistFragment.java
```diff
@@ -253,7 +253,7 @@ else if (entry.isPlayed()) {
                 element.findViewById(R.id.now_playing).setVisibility(View.INVISIBLE);
             } 
             else {
-                element.setBackgroundColor(getResources().getColor(com.actionbarsherlock.R.color.abs__background_holo_light));
+//                element.setBackgroundColor(getResources().getColor(com.actionbarsherlock.R.color.abs__background_holo_light));
```
theJenix: TODO: needs fixing or removal/comment.

theJenix:
All in all this looks like a pretty straightforward refactor (provided the code behind the scenes works about the same).  There are a few TODOs, and it looks like some styles or style methods aren't defined in AppCompat, but given that AppCompat is Google's supported library for the action bar, my guess is this will be pretty easy to validate once it's done.

SoundStream/res/values/colors.xml
reidbaker, line 37: Probably but lets have that be another pull request

SoundStream/src/com/thelastcrusade/soundstream/CoreActivity.java
reidbaker: This is why I shouldnt try an do a refactor in multiple sittings.

SoundStream/src/com/thelastcrusade/soundstream/components/PlaylistFragment.java
reidbaker: should be fixed now

PR #330, opened 2014-01-08
Cancel partial

### LINKED ISSUE #182: Removing partially transferred songs does not cancel the transfer
Observed in logcat on the host device.  The transfer runs to completion even if the song is not in the playlist anymore.

theJenix:
Cancel transfers when transfering songs are removed from the playlist.

Fixes #182


dgreenhalgh:
There's a lot here. I'll be looking through it today, but could we get some more eyeballs on it?

reidbaker:
There is a lot here. Maybe jesse can walk us through tonight

SoundStream/.classpath
```diff
@@ -5,8 +5,5 @@
 	<classpathentry kind="src" path="src"/>
 	<classpathentry kind="src" path="gen"/>
 	<classpathentry exported="true" kind="con" path="com.android.ide.eclipse.adt.DEPENDENCIES"/>
-	<classpathentry kind="src" path="/SlidingMenuLibrary"/>
-	<classpathentry kind="con" path="com.android.ide.eclipse.adt.ANDROID_FRAMEWORK/android-support-v7-appcompat"/>
```
reidbaker, line 9: These probably shouldnt be removed

SoundStream/src/com/thelastcrusade/soundstream/net/MessageEnqueuer.java
```diff
@@ -0,0 +1,59 @@
...
+/**
+ * @author Jesse Rosalia
+ *
+ */
+public class MessageEnqueuer {
+
+    private int mmOutMessageNumber = 1;
+    private ConnectionWriter writer;
+
+    /**
```
reidbaker: empty java doc

SoundStream/src/com/thelastcrusade/soundstream/net/MessageReceiver.java
```diff
@@ -0,0 +1,60 @@
...
+/**
+ * @author Jesse Rosalia
+ *
+ */
+public class MessageReceiver {
+
+    private int mmInMessageNumber  = 1;
+    private Handler mmHandler;
+
+    /**
```
reidbaker, line 39: empty java doc

SoundStream/src/com/thelastcrusade/soundstream/net/wire/CancelInputStream.java
```diff
@@ -0,0 +1,27 @@
...
+ * You should have received a copy of the GNU General Public License
+ * along with SoundStream.  If not, see <http://www.gnu.org/licenses/>.
+ */
+package com.thelastcrusade.soundstream.net.wire;
+
+/**
+ * @author theJenix
+ *
+ */
+public class CancelInputStream {
```
theJenix: Delete me

SoundStream/src/com/thelastcrusade/soundstream/net/MessageFuture.java
```diff
@@ -0,0 +1,62 @@
...
+ */
+package com.thelastcrusade.soundstream.net;
+
+import java.io.IOException;
+
+/**
+ * @author Jesse Rosalia
+ *
+ */
+public abstract class MessageFuture {
```
theJenix, line 27: Implement java.util.concurrent.Future

theJenix, line 27: Actually, looking at the Future interface, I'm not so sure this makes sense.  Our future doesn't return anything, and will not be used with a scheduled executor (since we control the execution as part of the message send process).  Because of this, I think it would be confusing to implement Java's Future interface unless we changed our message queue to be based on the other java concurrent structures (which I'm not sure is a good idea).

SoundStream/src/com/thelastcrusade/soundstream/net/MessageReceiver.java
theJenix, line 39: :thumbsup:

SoundStream/src/com/thelastcrusade/soundstream/net/MessageEnqueuer.java
theJenix: :thumbsup:

SoundStream/.classpath
theJenix, line 9: :thumbsup:

SoundStream/src/com/thelastcrusade/soundstream/net/wire/CancelInputStream.java
theJenix: :thumbsup:

SoundStream/src/com/thelastcrusade/soundstream/net/ConnectionWriter.java
```diff
@@ -76,22 +91,33 @@ public int compare(QueueEntry lhs, QueueEntry rhs) {
         }
     });
 
+    private Set<Integer> canceled = new HashSet<Integer>();
```
dgreenhalgh, line 94: "cancelled"

SoundStream/src/com/thelastcrusade/soundstream/net/ConnectionWriter.java
```diff
@@ -130,6 +157,52 @@ public boolean canWrite() {
...
+                }
+            }
+            if (found != null) {
+                queue.remove(found);
+            }
+        }
+        return found;
+    }
+
+    private void processCanceled() throws IOException {
```
dgreenhalgh, line 184: sp

SoundStream/src/com/thelastcrusade/soundstream/net/ConnectionWriter.java
```diff
@@ -130,6 +157,52 @@ public boolean canWrite() {
...
+        Set<Integer> canceled = new HashSet<Integer>();
```
dgreenhalgh, line 185: :hamburger:

SoundStream/src/com/thelastcrusade/soundstream/net/MessageReceiver.java
```diff
@@ -0,0 +1,63 @@
...
+ * A helper class to receive messages from the wire and dispatch them through the Android
+ * system.
+ * 
+ * @author Jesse Rosalia
+ *
+ */
+public class MessageReceiver {
+
+    private int mmInMessageNumber  = 1;
+    private Handler mmHandler;
```
dgreenhalgh: One 'm' is probably sufficient to distinguish that these two are member variables

SoundStream/src/com/thelastcrusade/soundstream/net/bt/BluetoothConnection.java
```diff
@@ -0,0 +1,134 @@
...
+ * This thread is responsible for sending and receiving messages once the connection has been established.
+ * 
+ * 
+ * @author Jesse Rosalia
+ *
+ */
+public abstract class BluetoothConnection extends Thread {
+    private final String TAG = BluetoothConnection.class.getSimpleName();
+
+    private final BluetoothSocket mmSocket;
```
dgreenhalgh: Same with the prefix of all of these member variables, unless this is something that we've been doing that I just missed

SoundStream/src/com/thelastcrusade/soundstream/net/bt/BluetoothConnection.java
```diff
@@ -0,0 +1,134 @@
...
+        try {
+            mmSocket.close();
+        } catch (IOException e) {
+            
+        } finally {
+            //before we exit, notify the launcher thread that the connection is dead
+            onDisconnected();
+        }
+    }
+}
```
dgreenhalgh: newline

SoundStream/src/com/thelastcrusade/soundstream/service/TransferService.java
```diff
@@ -0,0 +1,205 @@
...
+
+    @Override
+    public IBinder onBind(Intent intent) {
+        return new TransferServiceBinder();
+    }
+
+
+    /**
+     * 
+     */
```
dgreenhalgh, line 95: Empty javadoc

SoundStream/src/com/thelastcrusade/soundstream/service/TransferService.java
```diff
@@ -0,0 +1,205 @@
...
+//                    String fromAddr = intent.getStringExtra(MessagingService.EXTRA_ADDRESS);
+//                    long   songId   = intent.getLongExtra(  MessagingService.EXTRA_SONG_ID, SongMetadata.UNKNOWN_SONG);
+//
+//                    if (songId == SongMetadata.UNKNOWN_SONG) {
+//                        Log.wtf(TAG, "REQUEST_SONG_MESSAGE action received without a valid song id");    
+//                    } else {
+//                        sendSongData(fromAddr, songId);
+//                    }
+//                }
+//            })
```
dgreenhalgh: What's the status on this block of code?

dgreenhalgh:
I had some spelling/formatting nitpicks, but overall I like this.

SoundStream/src/com/thelastcrusade/soundstream/net/MessageFuture.java
reidbaker, line 27: Ok I trust your call. @twrobel3 do you agree?

SoundStream/src/com/thelastcrusade/soundstream/net/ConnectionWriter.java
reidbaker, line 94: haha :panda_face:

SoundStream-test/src/com/thelastcrusade/soundstream/net/message/MessengerTest.java
```diff
@@ -53,6 +54,92 @@
     //TODO: add tests for partial messages, to make sure we handle the case where data isn't all there yet
     
     @Test
+    public void testCancelMessage() throws Exception {
```
reidbaker, line 57: LOVE that you wrote a test but there a bunch of hard coded numbers (mostly 0's) can these be put into variables?

SoundStream/src/com/thelastcrusade/soundstream/net/MessageFuture.java
taywrobel, line 27: If it doesn't make sense to have it extend future I'd at least like to see a renaming of our implementation, since anyone that looks at the code and is familiar with java concurrency will be confused by the naming conflict.

Moreover, why do you not think basing our queue on java concurrency practices and infrastructure is a good idea?

SoundStream-test/src/com/thelastcrusade/soundstream/net/message/MessengerTest.java
```diff
@@ -53,6 +54,92 @@
...
+        received = messenger.deserializeMessage(buffer.getInputStream());
+        assertFalse(received);
+
+        //should still have 0 active transfers and 0 received messages
+        assertEquals(0, messenger.getActiveTransferCount());
+        assertEquals(0, messenger.getReceivedMessages().size());
+    }
+
+    @Test
+    public void testDeserializeFileMessagePartialReceive() throws Exception {
```
reidbaker, line 105: Same for this test

SoundStream/.classpath
```diff
@@ -5,8 +5,8 @@
 	<classpathentry kind="src" path="src"/>
 	<classpathentry kind="src" path="gen"/>
 	<classpathentry exported="true" kind="con" path="com.android.ide.eclipse.adt.DEPENDENCIES"/>
-	<classpathentry kind="src" path="/SlidingMenuLibrary"/>
```
reidbaker, line 8: Not sure this file should be updated.

SoundStream/project.properties
```diff
@@ -12,5 +12,5 @@
 
 # Project target.
 target=android-17
-android.library.reference.1=../third_party/SlidingMenuLibrary
```
reidbaker, line 15: Not sure this file should be changed. I think this might break the rest of our builds.

SoundStream/src/com/thelastcrusade/soundstream/net/ConnectionReader.java
```diff
@@ -0,0 +1,63 @@
...
+ */
+public class ConnectionReader {
+
+    private InputStream inStream;
+    private Messenger messenger;
+    private String remoteAddress;
+
+    /**
+     * @param remoteDevice 
+     * @param mmMessenger
```
reidbaker: mm?

SoundStream-test/src/com/thelastcrusade/soundstream/net/message/MessengerTest.java
taywrobel, line 57: I disagree that there's a need to make variables for numbers in tests that are not special.  If anything we could make an "assert queue is empty" method that does the check with 0, but more than that I find just pedantic

SoundStream/src/com/thelastcrusade/soundstream/net/ConnectionWriteThread.java
```diff
@@ -0,0 +1,96 @@
...
+        writeThreadRunning = true;
+        super.start();
+    }
+    
+    public void stopAndWait() {
+        writeThreadRunning = false;
+        stoppingThread = Thread.currentThread();
+        synchronized(stoppingThread) {
+            try {
+                //wait for the thread to stop, or for 1 second.  This number may have to be adjusted
```
reidbaker: What might cause us to need to adjust this number? What are the side effects?

SoundStream/.classpath
taywrobel, line 8: It should not, this is IDE specific.  I should probably add .classpath and .project to the gitignore

SoundStream/src/com/thelastcrusade/soundstream/net/MessageEnqueuer.java
```diff
@@ -0,0 +1,64 @@
...
+ * 
+ * This class also defines the behavior of the Future and the cancel method,
+ * for canceling outstanding messages.
+ * 
+ * @author Jesse Rosalia
+ *
+ */
+public class MessageEnqueuer {
+
+    private int mmOutMessageNumber = 1;
```
reidbaker: mm?

SoundStream/project.properties
taywrobel, line 15: Same as with the .classpath file, I should add this to the git ignore

SoundStream/src/com/thelastcrusade/soundstream/net/bt/BluetoothConnection.java
```diff
@@ -0,0 +1,134 @@
...
+    }
+
+    public boolean isRemoteDevice(BluetoothDevice device) {
+        return mmSocket.getRemoteDevice().equals(device);
+    }
+
+    public void run() {
+        // Keep listening to the InputStream until an exception occurs
+        MessageReceiver receiver = new MessageReceiver(mmHandler);
+        while (true) { //TODO: need way to kill this thread normally
```
reidbaker: Is there a ticket to add this later? Or will it happen before this code gets pulled in?

taywrobel: There is really no good way to kill a thread peacefully.  The best is to have a volatile variable that stores the alive state, and a cancellation method that flips the variable.

SoundStream/src/com/thelastcrusade/soundstream/net/wire/Messenger.java
```diff
@@ -258,10 +294,53 @@ private int readNext(InputStream input) throws IOException {
...
+        Map<Integer, Long> toKeep = new HashMap<Integer, Long>();
+        for (Map.Entry<Integer, Long> entry : this.canceledMessages.entrySet()) {
+            if ((System.currentTimeMillis() - entry.getValue()) / msPerMin <= this.canceledMessagesTtlMinutes) {
+                toKeep.put(entry.getKey(), entry.getValue());
+            }
+        }
+        this.canceledMessages = toKeep;
+    }
+
     ///TODO
```
reidbaker: We should either label this todo or remove it.

SoundStream/src/com/thelastcrusade/soundstream/net/MessageThreadMessageDispatch.java
```diff
@@ -85,19 +85,23 @@ public void handleMessage(Message msg) {
...
         
-        if (msg.what == MessageThread.MESSAGE_READ) {
+        if (msg.what == ConnectionConstants.MESSAGE_READ) {
             messageNo = msg.arg1;
             message   = (IMessage)msg.obj;
-            fromAddr  = msg.getData().getString(MessageThread.EXTRA_ADDRESS);
+            fromAddr  = msg.getData().getString(ConnectionConstants.EXTRA_ADDRESS);
             handleMessage(messageNo, message, fromAddr);
+        } else if (msg.what == ConnectionConstants.MESSAGE_FINISHED) {
+            //TODO: NYI
```
taywrobel: I don't think we should be adding unimplemented cases to our code base

SoundStream/src/com/thelastcrusade/soundstream/service/TransferService.java
reidbaker: The status is commented. :lollipop: -

SoundStream/src/com/thelastcrusade/soundstream/util/InputBuffer.java
```diff
@@ -58,14 +58,18 @@ public int read(byte[] buffer, int offset, int length)
...
             int toCopy = Math.min(length, available());
             System.arraycopy(buf, index, buffer, offset, toCopy);
             index += toCopy;
             return toCopy;
         }
+        
         @Override
         public int read() throws IOException {
-            return index < size() ? buf[index++] : -1;
+            //read returns an integer between 0 and 255, or -1
```
reidbaker, line 71: :thumbsup:

SoundStream-test/src/com/thelastcrusade/soundstream/net/message/MessengerTest.java
reidbaker, line 57: It gives meaning to the numbers. For example we use 1 for a location then
use it later to check what we expect is in location 1. Making it a variable
let's the code be easier to read.

SoundStream/src/com/thelastcrusade/soundstream/net/MessageFuture.java
theJenix, line 27: I don't think people will be confused; future is not a java API specific term, and they can see that the class does not implement Java's Future.

Regarding the java concurrency infrastructure, I don't think it fits our model or our purposes.  An ExecutorService will, as I understand it, take n tasks and distribute them across k threads (predefined or launched as needed), and return Future objects as a way of obtaining the results/status at a later date.  Resource contention is resolved on a first come, first serve basis, which for most background tasks (e.g. searching, sorting, indexing, etc) is what you want.  In this case, the queue and surrounding class are responsible for transmitting one or more simultaneous messages from one device to another; we use a single thread so we can control who can "jump the queue", which allows us to optimize for fast transmission of songs but still allow control messages to be sent immediately.  In our case, first come, first serve is the wrong model for contention access (e.g. access to the socket).  I'm all for using standard structures when they make sense, and if I'm wrong in my understand or assumptions, please let me know, but I really think we're talking about two solutions to two different problems.

SoundStream/src/com/thelastcrusade/soundstream/net/ConnectionWriter.java
theJenix, line 94: Canceled is the accepted spelling in American English (ref: http://grammarist.com/spelling/cancel/).

SoundStream/src/com/thelastcrusade/soundstream/net/ConnectionWriter.java
theJenix, line 184: See above comment.  This spelling is correct.

SoundStream/src/com/thelastcrusade/soundstream/net/ConnectionWriter.java
theJenix, line 185: Above.

SoundStream/src/com/thelastcrusade/soundstream/net/bt/BluetoothConnection.java
theJenix: :thumbsup:

SoundStream/src/com/thelastcrusade/soundstream/net/MessageReceiver.java
theJenix: :thumbsup: I removed the mm from all member variables, which should be consistent with other parts of the code.  "mm" is from example code I pulled from Android docs.

SoundStream/src/com/thelastcrusade/soundstream/service/TransferService.java
theJenix, line 95: :thumbsup:

SoundStream/src/com/thelastcrusade/soundstream/service/TransferService.java
theJenix: aaaaaand it's gone.

SoundStream-test/src/com/thelastcrusade/soundstream/net/message/MessengerTest.java
theJenix, line 57: I agree with both of you guys in principle...for this code, though, it's more what Taylor is describing than what Reid is describing (e.g. the numbers tend to mean "empty" or "has one" or "has 2") and it's fairly well commented in most cases as to what they mean.  In other cases, where we're pulling out a specific position or a specific ID, variables or constants are definitely the way to go.  It would also be worth someone taking a look at all of our tests, and pulling out common patterns that we can make clearer, which should help us write more, better tests in the future.  Maybe that can be an action that we attach to our automated testing work?  (Not to say Taylor has to do it all... :) )

SoundStream/.classpath
theJenix, line 8: Once we start using gradle, I totally agree...until then though, I think not having these files will make it confusing for anyone just getting into the project.

SoundStream/project.properties
theJenix, line 15: Same caveat as above...I encourage us to transition to gradle before excusing Eclipse specific files.

SoundStream/src/com/thelastcrusade/soundstream/net/ConnectionWriteThread.java
theJenix: This method signals that the running thread defined by this class should terminate (by setting writeThreadRunning to false).  There's a loop in the run method that is conditional on this flag that calls writer.writeOne, then sleeps for 10 ms (if the writer indicates it cannot write); this loop is what actually writes packets out to the connected socket's OutputStream.  The wait here, therefore, is to wait for at most one iteration of that loop.  As long as writeOne doesn't take a long time to complete, a second is probably overkill....but writeOne takes longer as we up the # of bytes we send at a time.  I'll fix this comment because it's really just a warning for when we start messing with the max write size.

SoundStream/src/com/thelastcrusade/soundstream/net/MessageEnqueuer.java
theJenix: :thumbsup:

SoundStream/src/com/thelastcrusade/soundstream/net/ConnectionWriter.java
dgreenhalgh, line 94: I stand corrected and cancell my comment

SoundStream/src/com/thelastcrusade/soundstream/net/bt/BluetoothConnection.java
theJenix: This particular thread should stay alive as long as the connection is alive, OR terminate itself on an exception (hence the break on caught exceptions, and then the cleanup code).  I cleaned this up, though...added a local flag, and refactored the try/catch block, because I don't want to swallow any Throwables that ACRA should be handling.

SoundStream/src/com/thelastcrusade/soundstream/net/wire/Messenger.java
theJenix: Removed :thumbsup:

SoundStream/src/com/thelastcrusade/soundstream/net/MessageThreadMessageDispatch.java
theJenix: Agreed..it's gone.

SoundStream/src/com/thelastcrusade/soundstream/net/ConnectionWriter.java
theJenix, line 94: :us:

SoundStream/src/com/thelastcrusade/soundstream/net/bt/BluetoothConnection.java
theJenix: Added :thumbsup:

theJenix:
Just pushed a ton of changes that should address the spelling, comments and other smaller issues.  Regarding the testing and Future discussion...if we need to discuss further, I suggest we move to a Hangout (for efficiency's sake).

dgreenhalgh:
Hmm... I happen to know of a Hangout happening soon

SoundStream/src/com/thelastcrusade/soundstream/net/MessageFuture.java
taywrobel, line 27: I'll need to run into work to grab my JCIP book to verify, but I'm pretty sure we can do all this with the concurrency infra and both accomplish our goal and take advantage of the benefits.

A thread pool executor can be given a queue implementation for the tasks to implement.  If we make this a priority blocking queue then we can arbitrarially order tasks to be executed.  All we need to do is hand it ComparableFutureTasks that contain the connection task and are prioritized by queue position.

So, we could still have an executor manage the transferring, reducing resource usage, and adding stability, and, we can start with a thread pool of only 1 thread, matching the current model, but also have the ability to increase it trivially if desired (such as if we want to support concurrent transfers over wifi at some point in the future).

theJenix, line 27: The current implementation works (there's nothing inherently unstable about it) and changing that code will require a lot of testing, but I like your suggestion for if we ever want to support concurrent sends over a socket, so I propose a compromise.  Let's accept this for now, and create an issue tied to supporting a new communications method for us to revisit the threading implementation of our message queue.

dgreenhalgh:
:shipit:

PR #333, opened 2014-01-18
Remove the last bit of ABS from app

reidbaker:
App tested with single device and works. Text color in action bar is too light.


taywrobel:
Closed the wrong one, sorry.

dgreenhalgh:
Makes sense. We can consult @ejohnson44 on the text color at tonight's meeting.

theJenix:
:ship:

PR #335, opened 2014-01-28
Search improvements 

reidbaker:
search no longer crashes, is selected after pressing search, results counter added but not working

This is not ready to be pulled in but all the code here will eventually make it into the repo


ejohnson44:
Looks good to me

ejohnson44:
Actually I lied. I didn't realize that reuslts counter stuff was in here

reidbaker:
@ejohnson44 if you merged my stuff and have results working then I can abandon this and we can take your pull request.

ejohnson44:
Sounds like a plan. I need to do a bit of formatting but other than that it's ready

PR #336, opened 2014-02-04
Search counter

ejohnson44:
Got the search counter up and running - also includes Reid's fixes to some crashes


SoundStream/project.properties
```diff
@@ -12,4 +12,5 @@
 
 # Project target.
 target=android-17
-android.library.reference.1=../third_party/SlidingMenuLibrary
+android.library.reference.1=..\\third_party\\SlidingMenuLibrary
+android.library.reference.2=../../../../AppData/Local/Android/android-sdk/extras/android/support/v7/appcompat
```
dgreenhalgh: This came up in another pull request, but is this filepath Windows-specific?

SoundStream/src/com/thelastcrusade/soundstream/components/MusicLibraryFragment.java
```diff
@@ -63,6 +65,8 @@
     
     private MusicAdapter mMusicAdapter;
     
+    private View headerView;
```
dgreenhalgh: Nitpick, but a better name would be mHeaderView to reinforce the concept that this is a member variable

SoundStream/project.properties
ejohnson44: It definitely is, but I don't remember how to not include it but not have it disappear from my project.

reidbaker: git rm should remove it from git but not remove it locally.

dgreenhalgh:
:+1:

SoundStream/src/com/thelastcrusade/soundstream/components/MusicLibraryFragment.java
```diff
@@ -313,13 +350,13 @@ public void onClick(View v) {
         @Deprecated
         private void updateMusicFromLibrary() {
             this.updateMusic(getMusicLibraryFromService());
-            notifyDataSetChanged();
+//            notifyDataSetChanged();
```
reidbaker: Remove?

SoundStream/src/com/thelastcrusade/soundstream/components/MusicLibraryFragment.java
```diff
@@ -313,13 +350,13 @@ public void onClick(View v) {
...
         }
         
         private void updateMusicFromQuery(String query) {
             if (query != null) {
                 this.updateMusic(getMusicLibraryFromQuery(query));
-                notifyDataSetChanged();    
+//                notifyDataSetChanged();
```
reidbaker: Remove?

SoundStream/src/com/thelastcrusade/soundstream/components/MusicLibraryFragment.java
theJenix: Why was this removed?  Is this handled in the class updateMusic function?

SoundStream/src/com/thelastcrusade/soundstream/components/MusicLibraryFragment.java
theJenix: Same comment as above.

SoundStream/src/com/thelastcrusade/soundstream/CoreActivity.java
```diff
@@ -188,7 +190,14 @@ public boolean onCreateOptionsMenu(Menu menu) {
         
         if (searchView != null) {
             searchView.setSearchableInfo(searchManager.getSearchableInfo(new ComponentName(this, SearchActivity.class)));
-            searchView.setIconifiedByDefault(true);
+            searchView.setIconified(true);
+            searchView.setOnClickListener(new OnClickListener() {
+                @Override
+                public void onClick(View v) {
+                    v.setFocusable(true);
+                    v.requestFocusFromTouch();
```
theJenix, line 198: It seems like this shouldn't be necessary (i.e. using a click listener to change the focusability of a view and to request focus should be handled by the core, no?), but I presume it is for some reason.  Can you add a comment explaining what's going on here?

SoundStream/src/com/thelastcrusade/soundstream/components/MusicLibraryFragment.java
```diff
@@ -137,9 +141,11 @@ public View onCreateView(LayoutInflater inflater, ViewGroup container,
         
         super.onCreateView(inflater, container, savedInstanceState);
         View v = inflater.inflate(R.layout.list, container, false);
+ 
+        setListAdapter(null);
```
theJenix, line 146: What's going on with this line?  Are we trying to reuse functionality that isn't exactly right for this purpose, and would it be better to refactor to support this case?

SoundStream/src/com/thelastcrusade/soundstream/components/MusicLibraryFragment.java
```diff
@@ -163,12 +190,22 @@ public void onServiceBound() {
...
     @Override
     public void onStart() {
         super.onStart();
 //        ((CoreActivity)getActivity()).getTracker().sendView(TAG);
     }
+    
+    @Override
+    public void onActivityCreated(Bundle savedInstanceState) {
+        super.onActivityCreated(savedInstanceState);
+        if(mHeaderView != null){
```
theJenix, line 193: How can mHeaderView be null?  Can you please add a comment?

SoundStream/src/com/thelastcrusade/soundstream/components/MusicLibraryFragment.java
```diff
@@ -147,10 +153,31 @@ public View onCreateView(LayoutInflater inflater, ViewGroup container,
...
+                mHeaderView = inflater.inflate(R.layout.search_counter,null);
+                mHeaderView.setLayoutParams(new ListView.LayoutParams(ListView.LayoutParams.MATCH_PARENT, ListView.LayoutParams.WRAP_CONTENT));
+                
+
+                mMusicAdapter.registerDataSetObserver(new DataSetObserver() {
+                    @Override
+                    public void onChanged() {
+                      super.onChanged();
+                      TextView resultsCounter = (TextView)mHeaderView.findViewById(R.id.results_count);
+//                      Log.i(TAG, "Music Adapter Count on observed change" + mMusicAdapter.getCount());
```
theJenix: You can wrap these with Log.isLoggable(TAG, Log.INFO) and leave these log statements in place, and we can enable that level in logcat if we need that level of detail.  Also, if appropriate, you LogUtil.isLogAvailable to make this test junit testable (if it's more appropriate to use the integration tests, I don't think we need to check.  This is just for straight up unit tests).

SoundStream/src/com/thelastcrusade/soundstream/components/MusicLibraryFragment.java
```diff
@@ -147,10 +153,31 @@ public View onCreateView(LayoutInflater inflater, ViewGroup container,
...
+                    }
+                });
+                
+            
             } else {
                 Log.w(TAG, "Fragment recieved arguments but no query");
             }
         }
+        else{
+            mQuery = null;
```
theJenix, line 174: This is a small nit, but an alternative to using null would be to use the empty string (or an appropriately named constant), or some other string to indicate all results.  This would clean up the test for nulls, and may streamline the actual search (i.e. if all queries were valid, including "no query", you wouldn't need any special cases for no query).  This is not a showstopper, more just a suggestion for any future changes.

SoundStream/src/com/thelastcrusade/soundstream/components/MusicLibraryFragment.java
ejohnson44: Yeah, it's done in another function so it's not necessary here. I'll get rid of the commented bit

SoundStream/src/com/thelastcrusade/soundstream/CoreActivity.java
ejohnson44, line 198: This is some of what Reid did, so I'll let him tell me what comment should go there

SoundStream/src/com/thelastcrusade/soundstream/components/MusicLibraryFragment.java
ejohnson44, line 146: It's some of what I had to do to make this 2.3 compatible. If we actually change away from that then I can get rid of this

SoundStream/src/com/thelastcrusade/soundstream/components/MusicLibraryFragment.java
ejohnson44: That was just for my own experimentation I think - it's not something that is necessary

SoundStream/src/com/thelastcrusade/soundstream/CoreActivity.java
reidbaker, line 198: this makes it so when you tap on the search icon in addition to opening the text box it selects it.

SoundStream/src/com/thelastcrusade/soundstream/components/MusicLibraryFragment.java
reidbaker, line 174: Like that solution

SoundStream/src/com/thelastcrusade/soundstream/CoreActivity.java
reidbaker, line 198: //Make search box selected after tapping search icon

reidbaker:
Is this good to pull in?

reidbaker:
@ejohnson44 says this can be pulled in

